PR #3, opened 2014-10-17
Closes #2, Allows paths to be excluded and included.

### LINKED ISSUE #2: Support file exclusion/inclusion
When I run `asar pack . app.asar` the result is 75MB! It must be including `node_modules/atom-package-manager` and when I `asar list app.asar` it looks like it's even bundled up `.git` — so much for using this to "protect your app's resources and source code from the users" like ill-advisedly [claimed](https://github.com/atom/atom-shell/blob/master/docs/tutorial/application-packaging.md)!

Can an option be provided to exclude directories, or manually list files to bundle?

larzconwell:


zcbenz:
@larzconwell Very sorry I just found this PR, can you rebase it on master branch?

bwin:
This uses `glob.sync`. I would opt for `minimatch`.

Just my 2 cents...

larzconwell:
I agree @bwin. I don't really write Node anymore so I'm just gonna close this PR.

IIRC I opted for `glob.sync` at the because `createPackage` didn't have a callback at the time and I didn't want to break api.

bwin:
Thanks for contributing anyway.

PR #19, opened 2015-01-26
fix extracting zero-size files

joshuawarner32:
Apparently, both node 0.10.35 and 0.11.14 have a bug (arguably) where they consider zero-size writes into a zero-size buffer to be invalid.  Intuitively, a zero-size write should always be a no-op (with no error), as long as the buffer is valid.

Previously, I was seeing this error when extracting an archive with a zero-size file:
```
fs.js:504
  var r = binding.read(fd, buffer, offset, length, position);
                  ^
Error: Offset is out of bounds
    at Error (native)
    at Object.fs.readSync (fs.js:504:19)
    at Object.module.exports.readFile (/Users/joshua.warner/myproject/node_modules/asar/lib/disk.js:64:6)
    at Object.module.exports.extractAll (/Users/joshua.warner/myproject/node_modules/asar/lib/asar.js:111:22)
    at Command.<anonymous> (/Users/joshua.warner/myproject/node_modules/asar/bin/asar:34:15)
    at Command.listener (/Users/joshua.warner/myproject/node_modules/asar/node_modules/commander/index.js:249:8)
    at Command.emit (events.js:110:17)
    at Command.parseArgs (/Users/joshua.warner/myproject/node_modules/asar/node_modules/commander/index.js:480:12)
    at Command.parse (/Users/joshua.warner/myproject/node_modules/asar/node_modules/commander/index.js:372:21)
    at Object.<anonymous> (/Users/joshua.warner/myproject/node_modules/asar/bin/asar:37:9)
```


lib/disk.js
```diff
@@ -57,10 +57,12 @@ module.exports.readFilesystem = function(archive) {
 };
 
 module.exports.readFile = function(filesystem, info) {
-  var fd = fs.openSync(filesystem.src, 'r');
   var buffer = new Buffer(info.size);
-  fs.readSync(fd, buffer, 0, info.size,
-              8 + filesystem.headerSize + parseInt(info.offset));
-  fs.closeSync(fd);
+  if(info.size > 0) {
```
zcbenz: Needs a space here: `if(` => `if (`.

And can you add some comment of the purpose in the code?

joshuawarner32:
@zcbenz - fixed.

zcbenz:
Thanks!

zcbenz:
asar@0.2.2 has been published for this fix.

PR #22, opened 2015-03-09
drop chromium-pickle dependency. implement reading / writing asar archives in pure JS

YuriSolovyov:
As follow-up to https://github.com/atom/asar/issues/16

### DIFF
--- a/lib/header.js
+++ b/lib/header.js
@@ -1,0 +1,48 @@
+var fs = require('fs');
+var pickleHeader = new Buffer([0x04, 0x00, 0x00, 0x00, 0xfc, 0x00, 0x00, 0x00, 0xf8, 0x00, 0x00, 0x00]);
+var headerTerminator = new Buffer([0x00, 0x00]);
+var pickleSizeOffset = -4;
+
+var getHeaderSize = function(fd) {
+  var sizeBuf = new Buffer(16);
+
+  if (fs.readSync(fd, sizeBuf, 0, 16, 0) !== 16)
+    throw new Error('Unable to read header size');
+
+  return sizeBuf.slice(pickleSizeOffset).readUInt32LE(0);
+};
+
+var getHeaderContent = function(fd, size) {
+  var headerBuf = new Buffer(size);
+  if (fs.readSync(fd, headerBuf, 0, size, 16) !== size)
+    throw new Error('Unable to read header');
+
+  return headerBuf.toString('utf8');
+};
+
+var writeHeader = function(dest, headerBuffer, sizeBuffer, callback) {
+  dest.write(pickleHeader);
+  dest.write(sizeBuffer);
+  dest.write(headerBuffer);
+  dest.write(headerTerminator, callback);
+};
+
+exports.read = function(archive) {
+  var fd = fs.openSync(archive, 'r');
+
+  var size = getHeaderSize(fd);
+  var content = getHeaderContent(fd, size);
+  fs.closeSync(fd);
+
+  return { content: content, size: size };
+};
+
+exports.write = function(dest, header, callback) {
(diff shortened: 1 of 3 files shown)


YuriSolovyov:
investigating failure

YuriSolovyov:
works even with 0.10 now, not sure if it is needed though. /cc @zcbenz

zcbenz:
Can you maybe move the operations of reading/writing to a new class or module? In the original code it is plain to see that we have written two Pickle objects into file, but in this PR they become confusing operations of `Buffer` and magic numbers.

lib/header.js
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error('Unable to read header');
+
+  return headerBuf.toString('utf8');
+};
+
+var writeHeader = function(dest, headerBuffer, sizeBuffer, callback) {
+  dest.write(pickleHeader);
+  dest.write(sizeBuffer);
+  dest.write(headerBuffer);
+  dest.write(headerTerminator, callback);
```
YuriSolovyov: maybe it is better to do
```javascript
  var finalBuffer = Buffer.concat([
      pickleHeader,
      sizeBuffer,
      headerBuffer,
      headerTerminator
  ]);
  dest.write(finalBuffer, callback);
```
here?

YuriSolovyov:
moved related code to own file, just have one question.

lib/header.js
bwin: IMHO this `Buffer.concat` just makes an unnecessary copy.

zcbenz:
It looks good to me, thanks!

zcbenz:
asar@0.3.0 has been published for this change.

YuriSolovyov:
@zcbenz can you confirm it works on windows?
It fails to launch an app built with new version for me, and I can't even investigate it till morning.

YuriSolovyov:
@zcbenz even though test pass, it does not work on linux either. shell pickle fails to read header string
```
user@y:~/fix-asar$ node_modules/.bin/asar l app.asar
/home/user/fix-asar/node_modules/asar/lib/disk.js:47
  var header = headerPickle.createIterator().readString();
                                             ^
TypeError: Data can not be extracted
    at TypeError (native)
    at Object.module.exports.readArchiveHeader (/home/user/fix-asar/node_modules/asar/lib/disk.js:47:46)
    at Object.module.exports.readFilesystem (/home/user/fix-asar/node_modules/asar/lib/disk.js:52:21)
    at Object.module.exports.listPackage (/home/user/fix-asar/node_modules/asar/lib/asar.js:67:15)
    at Command.<anonymous> (/home/user/fix-asar/node_modules/asar/bin/asar:17:23)
    at Command.listener (/home/user/fix-asar/node_modules/asar/node_modules/commander/index.js:249:8)
    at Command.emit (events.js:110:17)
    at Command.parseArgs (/home/user/fix-asar/node_modules/asar/node_modules/commander/index.js:480:12)
    at Command.parse (/home/user/fix-asar/node_modules/asar/node_modules/commander/index.js:372:21)
    at Object.<anonymous> (/home/user/fix-asar/node_modules/asar/bin/asar:37:9)

```
my bad. I should've tried it on simple app, and not rely on test passing only.

YuriSolovyov:
looks like it might be something with hardcoded pickle header
```
user@y:~/fix-asar$ xxd app.asar > b1.hex
user@y:~/fix-asar$ xxd app-new.asar > b2.hex
user@y:~/fix-asar$ diff b1.hex b2.hex
1c1
< 0000000: 0400 0000 d00a 0000 cc0a 0000 c60a 0000  ................
---
> 0000000: 0400 0000 fc00 0000 f800 0000 c60a 0000  ................
```

Issue #16, opened 2014-12-21
Changes to the asar format? Magic number, checksum, filesize


bwin:
Currently an asar archive consists of these parts:

```
HEADERSIZE HEADER FILES
```

I would like to add a magic number(/string) to the beginning like 'ASAR' or something similar.
This would be backwards incompatible, but we could read it in the old format if it doesn't start with the magic string and show a deprecation notice for now.
Why? Currently we would read any file, interpret the first 8 bytes as UINT64 and try to read that much bytes and try to interpret that as json. Although this works, IMHO a magic number (or file signature) would be a nicer way.

If you're (@zcbenz) ok with the idea, how about adding a checksum too? (When you do incompatible changes do them all at once.) In normal operation the checksum can be ignored, but it could be used to check the integrity of the file when someone wants to.
It would also be possible to put the checksum into the json-header, but I don't like that.

I'm also in favor of #4, putting the filesize at the end. (This one doesn't break anything.)

BTW is there a reason that the header (json-filelist) gets pickled and not just written to the file? Is this faster or better in any way?

I'm also aware, that these changes need to be reflected in atom-shell to have any meaning at all.

It could look like this:

```
MAGIC HEADERSIZE CHECKSUM HEADER FILES FILESIZE
```

The position of HEADERSIZE and CHECKSUM could also be switched, I don't care.

bwin:
@zcbenz Ok, just a quick question: What about ditching asar and using zip files instead?
I think the answer is no, because of #673 (asar for runtime-mode). Is it?

bwin:
I changed my mind (after playing around with bwin/asar-util) about a few things.

* I would put the header *behind* the payload. This makes changing an archive easier/possible. Also appending is simple (overwrite current header with new payload and new header).
Also this makes compressing files in one pass possible (because with the header in front, you need to know the compressed payload-file-size up front).
* I would put a `version` field in the root of the json header. This is the file format version.
* header-size should IMHO be an UINT32. I think json header > 4.2 GB is unrealistic. (otherwise UINT64)
* max-file-size and max-archive-size is 1<<53 (max-header-size is 1<<32)
* I'm not really sure about the checksum-part. You (Cheng) said, you want the checksum in json, but I don't like that. IMO generating this checksum is an uneccessary PITA because you have to null the checksum entry, to avoid including the checksum in generating the checksum. (I wrote that in a confusing way, admitted.)
I don't really like my version either, because you're stuck with a fixed checksum size.

How to parse?
[...]

PR #23, opened 2015-03-13
Revert "drop chromium-pickle dependency. implement reading / writing asa...

YuriSolovyov:
Reverts atom/asar#22
Although asar test itself are passing it completly breaks reading of asar package by atom-shell itself.
Let's revert it and find better solution.


zcbenz:
I think we should rewrite chromium-pickle in JavaScript instead of hacking on Pickle's headers.

YuriSolovyov:
well, I agree, but my skills end in js level :)
simpliest layout IMO could be
```
SIZE <- UInt32
HEADER <- UTF-8 JSON string
FILE-1 bytes
...
FILE-N bytes
```
this could be written in pure js, but we need to acess this in c++ too.
Is there way to rewrite asar support in atom-shell in pure js?

bwin:
> Is there way to rewrite asar support in atom-shell in pure js?

I tried exacly that, but ATM I think it's not possible since we need to support asar in all situations (e.g. protocol).

> I think we should rewrite chromium-pickle in JavaScript instead of hacking on Pickle's headers.

About that, my opinion stays the same. I would really like to drop pickle (and cuint BTW) completely.
We just need it to read/write uints and a string. We only operate on little endian platforms (correct me, if I'm wrong), so no need for further abstractiion of uints on the c/c++ side. And about the string, there was no need (IMHO) for "pickling" in the first place.

YuriSolovyov:
@zcbenz  Can we move discussion to a new issue?

bwin:
Move it back to #16.

PR #29, opened 2015-04-24
Use glob instead of walkdir

YuriSolovyov:
Fixes https://github.com/atom/asar/issues/24

### DIFF
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -2,38 +2,12 @@ var fs = require('fs');
 var path = require('path');
 var os = require('os');
 
-var walkdir = require('walkdir');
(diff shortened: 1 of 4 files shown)


YuriSolovyov:
Inspired by https://github.com/bwin/asar/commit/308d5ca

YuriSolovyov:
should I keep walkdir for tests or make test use glob too?

YuriSolovyov:
@zcbenz 
I moved crawler function into own file to de-dublicate things and re-used it in tests.

zcbenz:
Looks good to me, thanks!

PR #41, opened 2015-08-25
Fix #40: Add option "dot" to include hidden files

### LINKED ISSUE #40: Option to include hidden files
Currently, hidden files (starting with a dot) are ignored: https://github.com/atom/asar/blob/master/src/crawlfs.coffee#L6 and https://github.com/isaacs/node-glob#dots

It'd be great to have an option to include hidden files, for example `{ dot: true }`, this option can be passed directly to glob.

SamyPesse:
This PR fixes #40 and adds an option `dot` to `asar.createPackageWithOptions` to include files starting with `.` (aka "hidden files").

This option is also available from the command line using option `--hidden`.

This PR also updates unit tests.


thomasjo:
This looks good to me.

@zcbenz Looks like this is blocking https://github.com/atom/atom/pull/7877 by causing the `.ctags` file in symbols-view to not being seen by asar.

zcbenz:
I think we probably include hidden files by default, it actually surprised me that it didn't.

And maybe we don't even need to provide an option for that, tools like `zip` and `tar` always include hidden files.

SamyPesse:
:+1:  I agree that hidden files should be included by default, I thought there was a reason why hidden files were ignored in Atom builds. You should just pass the option `{ dot: true }` to glob.

zcbenz:
> I agree that hidden files should be included by default, I thought there was a reason why hidden files were ignored in Atom builds. You should just pass the option { dot: true } to glob.

Can you change this PR to do that?

SamyPesse:
@zcbenz I pushed an update: hidden files are included by default, by the option `dot` is still available to exclude hidden files:

```js
asar.createPackageWithOptions('./input', './output.asar', {
    dot: false
}, function() { });
```

or using command line with `--exclude-hidden` option:

```
$ asar p ./input ./output.asar --exclude-hidden
```

Let me know if that's good for you, or if you prefer to completely remove the `dot` option.

SamyPesse:
@zcbenz Have you had a chance to take a look at it yet ?

It's blocking us for release.

zcbenz:
It looks good to me, thanks!

zcbenz:
asar@0.8.0 has been published with the fix.

SamyPesse:
Thanks :beers:

PR #43, opened 2015-10-25
Add 'unpack-dir' option to unpack specific directory

hokein:
The `unpack` option is for unpacking files. In some scenarios, it's not suffice. For example(part of maxogden/electron-packager#114), when packaging an app using a flash plugin to `asar` on OS X, there is no way to unpack `PepperFlashPlayer.plugin` since the flash plugin is a directory on OS X. So with this patch, we can easily make the asar app work well with flash plugin.


zcbenz:
:+1:

PR #44, opened 2015-10-26
Improve 'unpack-dir' option

hokein:
After more test, there are still some following works in #43. With the patch, the 'unpack-dir' is enough to work well:

* Make it work with 'unpack' option.
* Support unpack having-subdir directories.
* Improve testcase.


zcbenz:
:+1:

nantas:
Hi @hokein 

Thank you for making this option. However I got errors when using 'unpackDir' option to exclude 'node_modules' folder from being packed:

```js
/Users/nantas/projects/fireball-x/fb/node_modules/asar/lib/disk.js:41
        return callback(error);
               ^

RangeError: Maximum call stack size exceeded
    at writeFileListToStream (/Users/nantas/projects/fireball-x/fb/node_modules/asar/lib/disk.js:41:16)
    at writeFileListToStream (/Users/nantas/projects/fireball-x/fb/node_modules/asar/lib/disk.js:43:14)
    at writeFileListToStream (/Users/nantas/projects/fireball-x/fb/node_modules/asar/lib/disk.js:43:14)
```

The `return callback(error);` has been called too many times. When I try to log the acutal error I only get a very vague line:

```js
util.js:15
      objects.push(inspect(arguments[i]));
```

which I believe is from here: https://github.com/defunctzombie/node-util/blob/master/util.js#L27

I'm not sure why this line would pop an error. I'll do more test including hack the way copy files works in asar.

PR #52, opened 2015-12-22
Add support for a hint file to optimize archive ordering

bengotow:
Overview here: https://github.com/atom/atom/issues/10163


zcbenz:
:+1:

PR #54, opened 2016-01-23
Change "--unpack-dir" to use glob pattern

hmak3d:
Previously, `--unpack-dir` only did prefix match against literal string.  Coupled with [commander limitation that options can only occur once](https://github.com/tj/commander.js/blob/v2.9.0/index.js#L411), this means that you cannot exclude more than one directory from asar packing.

This patch makes `--unpack-dir` behave more like `--unpack` and accept glob patterns ... so you can specify more than one directory using brace expansion (see [minimatch](https://www.npmjs.com/package/minimatch)).

For backwards compatibility, `--unpack-dir` will still do literal prefix match \[if there's no glob match\].


zcbenz:
:+1:

PR #57, opened 2016-02-09
Add new method `createPackageFromFiles` to asar module

bwin:
Hi Cheng, haven't worked on (or with asar) for quite some time. But I still have a grunt and a gulp plugin for it on npm which need to be updated *badly*.

This is just another method to create an archive. This is helpful for grunt and gulp, when some other code already did the globbing.

The code didn't change that much, as GitHub makes it seem. Most lines have just one level less indentation.


bwin:
If this gets merged, I'll throw in a test (either in js or coffee, depending on #58).

zcbenz:
:+1:

PR #58, opened 2016-02-09
Rewrite tests in coffeescript

bwin:
- Tests in coffeescript -> yay!
- Travis CI: also run tests on node 4.x & 5.x
- updated dependencies

**Attention:** I changed all dependency versions to caret ranges (except `grunt-coffeelint`). I didn't check the changelog for all of them and there are some big version jumps. But our tests pass, so I assumed it's alright.
I had to update `minimatch` for some reason to pass the tests on node 5.

What's the reason for the pinned `grunt-coffeelint`?


zcbenz:
Thanks!

PR #64, opened 2016-04-29
Requires '-g' to use `asar` from Bash.

izy521:


izy521:
Actually,  I don't think this requires a PR, I'll just drop an issue.

PR #68, opened 2016-05-24
fix: escape src dir while crawling fs

demetris-manikas:
Solves situations where the directory name is like "pa[ck]this"


develar:
glob is not a regexp. So, using `escape-string-regexp` is not fully correct.

Also, what electron-packager what — just ask pack the whole dir. It is not expected that path will be as a pattern.

So, I don't like current solution — we don't need to fight against asar/glob. Instead, should be a way to disable glob. And use path just as path, not as a pattern.

develar:
Issue: https://github.com/electron-userland/electron-builder/issues/371#issuecomment-221262908

PR #73, opened 2016-07-09
Adds transform option to api

gerhardberger:
This PR adds a `transform` option to the api, which is a function, that can return a Transform stream for files that needed to be transformed before copying them into the `.asar` file.


src/filesystem.coffee
```diff
@@ -28,23 +29,47 @@ class Filesystem
...
-      node.size = stat.size
+      node.size = file.stat.size
       node.unpacked = true
+      callback()
       return
 
-    # JavaScript can not precisely present integers >= UINT32_MAX.
-    if stat.size > 4294967295
-      throw new Error("#{p}: file size can not be larger than 4.2GB")
+    self = this
```
zcbenz: You can use the `=>` function in CoffeeScript, which handles this automatically.

gerhardberger: thanks I changed it!

zcbenz:
👍

PR #87, opened 2016-09-12
Make the CoffeeScript to ES6 JavaScript.

duncup:
Use ES6 JavaScript review src folder.
All test files no change.
Delete src folder.
New file all in the lib folder.
Delete `Gruntfile.coffee` and `coffeelint.json`.
Change .gitignore to remove the `/lib` form ignore list.
Clean package.json.


duncup:
I see the checks have old version check.
We need make old version checks be passed?

.travis.yml
```diff
@@ -1,9 +1,8 @@
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
```
kevinsawicki: I think these versions of node should still be supported at least through the end of 2016 since they are still valid LTS versions.

kevinsawicki:
Thanks for this pull request, it would be great to keep support for node 0.10 and 0.12 until they are officially end of lifed.

To do this, you could start by just compiling to JavaScript using CoffeeScript and check in the generated files which should be compatible with node 0.10/0.12.

That would remove the CoffeeScript and Grunt dependencies, and then in the future we could come back and add newer ES6 usage like `let`, `const`, etc. once we are ready to drop support for older node versions.

.travis.yml
```diff
@@ -1,9 +1,8 @@
...
   - "4"
   - "5"
+  - "6"
```
kevinsawicki: 👍 Adding test coverage for node 6 sounds good.

duncup:
So we need CoffeeScript befor LTS expired?
Get this.
I'll request again after `0.10` and `0.12` LTS expired.

PR #63, opened 2016-04-25
UnpackDir=xyz/{123,abc} pattern didn't match UnpackDir=xyz/123 and UnpackDir=xyz/abc combined

MathijsvVelde:
UnpackDir=xyz/{123,abc} pattern now matches UnpackDir=xyz/123 and UnpackDir=xyz/abc combined

PR #99, opened 2016-12-01
Fix issue with unpack-dir glob matching

kevinsawicki:
This is a continuation of https://github.com/electron/asar/pull/63 with added specs.

/cc @MarshallOfSound


MarshallOfSound:
My knowledge of CoffeeScript is limited but it looks good from here 😆

Issue #106, opened 2017-02-03
Intermittent truncated file content


njentit:
Listening on transforn end [https://github.com/electron/asar/blob/master/lib/filesystem.js#L84](url) will producing intermittent output as writer stream maybe not yet closed even when transfrom has reached end state.  I suggest change it into listening on writer close:

`return out.on('close', function() {`

For my case it producing consistent output.

pauliusuza:
I have a use case with electron-builder since version v14.2.0 (it's working in v14.1.1).


```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /repo/node_modules/electron-builder/src/util/readPackageJson.ts:76:46
    at Generator.throw (<anonymous>)
    at Generator.tryCatcher (/repo/node_modules/bluebird/js/release/util.js:16:23)
```

develar:
Electron-builder will use own old forked version, sorry for inconvenience, I will check what's wrong with latest asar.

malept:
It's unclear whether this is still a bug in the current version of `asar`. If it is, please comment with a way to reliably reproduce this (so we can figure out how to write a regression test in the testsuite) and we can reopen this issue.

PR #113, opened 2017-03-04
Do not return before write stream fully closes

pauliusuza:
Related to #106

### DIFF
--- a/lib/filesystem.js
+++ b/lib/filesystem.js
@@ -81,7 +81,7 @@ class Filesystem {
         const stream = fs.createReadStream(p)
 
         stream.pipe(tr).pipe(out)
-        return tr.on('end', function () {
+        return out.on('close', function () {
           file.transformed = {
             path,
             stat: fs.lstatSync(path)


lib/filesystem.js
```diff
@@ -81,7 +81,7 @@ class Filesystem {
         const stream = fs.createReadStream(p)
 
         stream.pipe(tr).pipe(out)
-        return tr.on('end', function () {
```
develar: `finish` should be used.

develar:
@pauliusuza this issue is not related to electron-builder issue because transform is not used by electron-builder.

lib/filesystem.js
develar: I am wrong, "close" is correct, https://github.com/jprichardson/node-fs-extra/issues/326

fantasydr:
I tried to use custom transform and got unstable result. It seems the stream is not closed properly.
This PR fixed the issue.
Code to repro:
```
var asar = require('asar');
var Transform = require('stream').Transform;
var fs = require('fs');

class CustomTransform extends Transform {
    _transform(chunk, enc, cb) {
        console.log(chunk.length, enc, cb); // do nothing, just print logs
        this.push(chunk);
        cb();
    };
}

function transform(filename) {
    console.log(filename);
    return new CustomTransform();
}

var src = 'input/';
var dest = 'app.asar';
asar.createPackageWithOptions(
    src, dest, 
    { transform }, 
    function() { 
        console.log('done.');
    }
)
```

njbmartin:
Surprised this hasn't been merged yet. This fixes the issues we've had transforming multiple files.

benlancaster:
👍  This bug is driving me nuts

pauliusuza:
Thanks for merging 👍

PR #117, opened 2017-03-29
Support passing an array of transform functions

benlancaster:
As a developer, I would like to pass my files through one *or more* transform functions when files are packed. This PR enables that and is backwards compatible with current behaviour.

Usage:

```javascript
var asar = require('asar');

var src = 'some/path/';
var dest = 'name.asar';

function firstTransform(filename) {
  return new FirstTransformStream()
}

function secondTransform(filename) {
  return new SecondTransformStream()
}

asar.createPackageWithOptions(src, dest, { transform: [firstTransform, secondTransform] }, function() {
  console.log('done.');
})

// single transforms still work too
asar.createPackageWithOptions(src, dest, { transform: firstTransform }, function() {
  console.log('done.');
})
```


lib/filesystem.js
```diff
@@ -73,22 +73,35 @@ class Filesystem {
...
-          file.transformed = {
-            path,
-            stat: fs.lstatSync(path)
-          }
+    if (options.transform instanceof Function) {
+      options.transform = [options.transform]
+    }
+
+    if (options.transform instanceof Array && options.transform.length) {
+      options.transform = options.transform.filter(() => true)
```
benlancaster: this re-indexes the array so we can deal with arrays with non-consecutive keys

MarshallOfSound: @benlancaster Would be cool if this is something new but as far as I can tell this is a noop call that simply clones the array.  Can you explain exactly what this fixes / handles?

lib/filesystem.js
```diff
@@ -73,22 +73,36 @@ class Filesystem {
...
+
+      const doTransform = (path, transformers, idx) => {
+        idx = idx || 0
+        const transform = transformers.hasOwnProperty(idx) ? transformers[idx](p) : false
```
MarshallOfSound: This seems a bit weird again, can you explain why need to check that `idx` exists

MarshallOfSound: And this should probably be `transformers[idx] ? transformers[idx](p) : false`

lib/filesystem.js
benlancaster: Take a look at the unit test I tweaked in a3154ad. It does clone the array but also re-orders the keys so if you have an array with non-contiguous keys, you end up with a an array with incrementing keys starting at zero. It's perhaps overly defensive

lib/filesystem.js
benlancaster: `hasOwnProperty` is in case the Array prototype has been messed with. Again, perhaps overly defensive. We check if there's a transform function to use at the start of doTransform, but you're right; we could just as easily check if there's another transform function at the end and only recurse if we need to

benlancaster: hasOwnProperty is in case the Array prototype has been messed with. Again, perhaps overly defensive. We check if there's a transform function to use at the start of doTransform, but you're right; we could just as easily check if there's another transform function at the end and only recurse if we need to

lib/filesystem.js
MarshallOfSound: @benlancaster Neat trick, would consider it a bit overly defensive though.  Passing arrays like that around would blow up most code in NPM 😆

benlancaster: I can take it out if you'd rather?

benlancaster:
Found some bugs. Will re-open when I get them fixed

PR #118, opened 2017-03-31
Making asar compatible with electron + adding functions to remove cache

arantes555:
Hello!

I'm building an Electron app and I needed to work with asar packages in it. I then ran in two problems.

First, the official `asar` package didn't work on electron (and the existing `original-fs-asar` uses old code and it does not look like it is maintained).

Second, as I was modifying asar packages, I had problems with the cache mechanism used by `asar`, which often ended up using the header of an asar to read a file in another asar, resulting in output from another file than the one I was reading (wrong offsets).

So, this pull request solves both these issues, making `asar` compatible with electron (of course while still being compatible with node), and adding two methodes to remove the cache.
I'd love to see those features merged into the official `asar`, as I'm sure they could be useful to other people! In the meantime, I published another package with these changes, under the name `asar-original-fs`.

Thank you !

### DIFF
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -216,3 +216,11 @@ module.exports.extractAll = function (archive, dest) {
     }
   })
 }
+
+module.exports.uncache = function (archive) {
+  return disk.uncacheFilesystem(archive)
+}
+
+module.exports.uncacheAll = function () {
+  disk.uncacheAll()
+}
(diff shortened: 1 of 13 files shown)


lib/asar.js
```diff
@@ -1,5 +1,5 @@
 'use strict'
-const fs = require('fs')
+const fs = process.versions.electron ? require('original-fs') : require('fs')
```
benlancaster: This is repeated in a number of places, is it worth moving this condition and require in to some kind of proxy module that returns either `original-fs` or `fs` depending on the environment?

arantes555: I don't believe it would be worth it. The condition is still pretty short & straightforward and can be inlined.

PR #121, opened 2017-05-18
Can't install asar as a dependancy of itself.

trilleplay:
Changed to regular npm install instead!

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ all files together without compression, while having random access support.
 ### Install
 
 ```bash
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
@trilleplay The intention with this section of the README is to instruct users to install `asar` as a dependency of their project or globally, these are usage instructions, not development instructions 👍

trilleplay:
Okay, Thanks for clarifying that! @MarshallOfSound

PR #127, opened 2017-08-28
Update disk.js to include read location

TJKoury:
Header position will always be at the beginning of file.  Explicitly stating position 0 guarantees the ready will happen from the start of the file, not the last read location on the [current file descriptor](https://nodejs.org/api/fs.html#fs_fs_read_fd_buffer_offset_length_position_callback).


TJKoury:
Having issues cross platform, will reopen when I solve them.

PR #154, opened 2018-08-17
make createPackageFromFiles works

akinoniku:
make createPackageFromFiles works without providing any options or metadata

calling like
`asar.createPackageFromFiles(toPath, fileName, paths, null, null, console.log)`
was not allowed.

### DIFF
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -62,6 +62,7 @@ callback: The callback function. Accepts (err).
 */
 module.exports.createPackageFromFiles = function (src, dest, filenames, metadata, options, callback) {
   if (typeof metadata === 'undefined' || metadata === null) { metadata = {} }
+  if (typeof options === 'undefined' || options === null) { options = {} }
   const filesystem = new Filesystem(src)
   const files = []
   const unpackDirs = []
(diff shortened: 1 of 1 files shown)


akinoniku:
Those test are not related to my changes, you don't have tests for `createPackageFromFiles`

zcbenz:
The tested methods are calling into `createPackageFromFiles` finally.

akinoniku:
ahh, ok, I will fix it

akinoniku:
Should work now, could you please review it, thanks.

lib/asar.js
```diff
@@ -103,7 +104,9 @@ module.exports.createPackageFromFiles = function (src, dest, filenames, metadata
 
     console.log(`Ordering file has ${((total - missing) / total) * 100}% coverage.`)
   } else {
-    filenamesSorted = filenames
+    for (const file of filenames) {
+      filenamesSorted.push(file.startsWith(src) ? file : path.join(src, file))
```
zcbenz: Hi, I'm removing this line since it is causing problems on Windows (https://github.com/electron/asar/issues/158).

MarshallOfSound: If someone wants to fix it the issue will be `/` and `\` (path separators)

PR #170, opened 2019-04-26
feat: drop support for Node < 8

malept:
BREAKING CHANGE: drop support for versions of Node which are EOL'd as of 2019-04-30.

The commit messages should give a pretty clear idea of what happened in this branch.


zcbenz:
I think we should release 2.0 for this change?

malept:
Hopefully #171 will cause this to happen automatically :smile: (assuming the `BREAKING CHANGE` is added to the squashed commit message)

PR #171, opened 2019-04-26
chore: add CFA support

malept:
I think this is all I have to do, code-wise?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     docker:
       - image: circleci/node:8
     steps:
-      - run: sudo apt-get install -y libnss3
+      - run: sudo apt-get install --no-install-recommends -y libgtk-3-0 libnss3
       - checkout
       - restore_cache:
           keys:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+  release:
+    docker:
+      - image: circleci/node:10.15
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package-lock.json" }}
+            - v1-dependencies-
+      - run: npm install
```
MarshallOfSound: Personally would prefer `yarn` (and it's locking strategy) over `npm` and a package lock.  That's the path I've been taking on other CircleCI + CFA conversions

zcbenz:
Do we have all the tokens set in Circle CI?

MarshallOfSound:
@zcbenz Yes I set up all the tokens we need :+1:

zcbenz:
Nice! I assume this should be merged before #170?

malept:
> I assume this should be merged before #170?

Yep!

PR #174, opened 2019-04-27
chore: use the correct version of the CFA module

malept:
Now with less spelling mistakes :grin:


malept:
Weird cache error in CircleCI...

PR #175, opened 2019-04-29
chore: have semantic-release update the CHANGELOG file

malept:
Does what it says on the tin.


malept:
Hrm. I missed the part in the docs where it actually gets committed to the git repo :grimacing:

PR #179, opened 2019-05-26
chore: multibyte characters path case use nosort

zp1996:
glob default sort result by function `localCompare`. Multibyte characters  path test case will failed when 
the system language is Chinese, so when run this case use nosort?


malept:
I think this is a chore rather than a fix, since it only affects tests.

zp1996:
> I think this is a chore rather than a fix, since it only affects tests.

I have changed the commit msg

PR #190, opened 2020-02-19
feat: add TypeScript definition

malept:
I made this mostly because I wanted to use it in Electron Packager's TypeScript definition, but it has the added bonus of partially documenting the API.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,65 @@
+import { Stats } from 'fs';
+
+export type CreateOptions = {
+  dot?: boolean;
+  globOptions?: {};
+  ordering?: string;
+  pattern?: string;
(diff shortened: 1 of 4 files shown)


lib/index.d.ts
```diff
@@ -0,0 +1,65 @@
+import { Stats } from 'fs';
+
+export type CreateOptions = {
+  dot?: boolean;
+  globOptions?: {};
```
HashimotoYT: Here should be a type for this? Options inside this object be any.

malept: That was an oversight, thanks for catching that. I'll need to add `@types/glob` to the dependencies.

malept: or possibly `optionalDependencies`?

PR #189, opened 2020-02-19
build: replace AppVeyor with CircleCI's Windows support

malept:
Consolidate CI providers now that CircleCI supports Windows.

PR #191, opened 2020-02-23
feat: require Node 10.12.0

malept:
BREAKING CHANGE: Drops support for Node < 10.12.0

This allows us to drop two direct dependencies (`cuint` and `mkdirp`), plus I found a way to drop a third (`tmp-promise`) that was unrelated to the Node version change.

Also uses `stream.pipeline` (added in Node 10) which simplifies how `Filesystem.prototype.insertFile` is implemented.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,13 +26,13 @@ version: 2.1
 orbs:
   win: circleci/windows@2.4.0
 jobs:
-  test-linux-8:
-    docker:
-      - image: circleci/node:8
-    <<: *steps-test
   test-linux-10:
     docker:
       - image: circleci/node:10
+    <<: *steps-test
+  test-linux-12:
+    docker:
+      - image: circleci/node:12
     <<: *steps-test
   test-mac:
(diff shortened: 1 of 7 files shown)


MarshallOfSound:
This unreasonably concerns me as I don't think the `asar` tests are solid enough for this kind of change. I'd like electron's test suite to run against this version of asar before we consider publishing it

malept:
> I'd like electron's test suite to run against this version of asar before we consider publishing it

Done: https://github.com/electron/electron/pull/22351

The only check that failed was Archaeologist, which can be fixed with https://github.com/electron/archaeologist/pull/16

malept:
Also want to point out that tests are passing, AppVeyor is a red herring due to https://github.com/electron/asar/pull/189#issuecomment-588426231

MarshallOfSound:
cc @zcbenz

zcbenz:
I changed Appveyor's setting to "Skip branches without appveyor.yml", so the appveyor errors should be gone in future.

PR #194, opened 2020-03-23
fix: correctly parse arguments for list --is-pack

### LINKED ISSUE #186: `asar list` may lose some output if it's writing to a pipe
`asar list example.asar | less` may not produce the complete output under Linux.

I believe `Process.exit(0)` should not be called after `console.log`.

See https://nodejs.org/api/process.html#process_process_exit_code and https://nodejs.org/api/process.html#process_a_note_on_process_i_o

malept:
build(deps): upgrade commander to ^5.0.0

Fixes #186.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "chromium-pickle-js": "^0.2.0",
-    "commander": "^4.1.1",
+    "commander": "^5.0.0",
     "glob": "^7.1.6",
     "minimatch": "^3.0.4"
   },
--- a/test/cli-spec.js
+++ b/test/cli-spec.js
@@ -12,10 +12,10 @@ const compDirs = require('./util/compareDirectories')
 const compFileLists = require('./util/compareFileLists')
 const compFiles = require('./util/compareFiles')
 
-childProcess.exec = promisify(childProcess.exec)
+const exec = promisify(childProcess.exec)
 
 async function execAsar (args) {
-  return childProcess.exec(`node bin/asar ${args}`)
+  return exec(`node bin/asar ${args}`)
 }
 
 async function assertAsarOutputMatches (args, expectedFilename) {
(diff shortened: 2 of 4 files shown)


test/cli-spec.js
```diff
@@ -118,7 +118,7 @@ describe('command line interface', function () {
     return assertAsarOutputMatches('l test/expected/packthis-unpack-dir.asar', 'test/expected/extractthis-filelist.txt')
   })
   it('should list files/dirs in archive with unpacked dirs & is-pack option', async () => {
-    return assertAsarOutputMatches('l test/expected/packthis-unpack-dir.asar --is-pack', 'test/expected/extractthis-filelist-with-option.txt')
+    return assertAsarOutputMatches('l --is-pack test/expected/packthis-unpack-dir.asar', 'test/expected/extractthis-filelist-with-option.txt')
```
malept: Swapping the position of the flag shows the bug with commander 4.x.

package.json
```diff
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "chromium-pickle-js": "^0.2.0",
-    "commander": "^4.1.1",
+    "commander": "^5.0.0",
     "glob": "^7.1.6",
     "minimatch": "^3.0.4"
```
Philhoy54, line 49: const { app, globalShortcut } = require('electron')

app.whenReady().then(() => {
  // Register a 'CommandOrControl+Y' shortcut listener.
  globalShortcut.register('CommandOrControl+Y', () => {
    // Do stuff when Y and either Command/Control is pressed.
  })
})

Philhoy54: Super

test/cli-spec.js
```diff
@@ -118,7 +118,7 @@ describe('command line interface', function () {
...
   })
   it('should extract an archive with unpacked dirs', async () => {
     await execAsar('e test/input/extractthis-unpack-dir.asar tmp/extractthis-unpack-dir/')
```
Philhoy54, line 124: @electron-bot @malept

Philhoy54:
image

PR #195, opened 2020-03-30
fix: CreateOptions.transform should return stream

toyobayashi:
According to [lib/filesystem.js#L65-L80](https://github.com/electron/asar/blob/3f243e2984b46991d9a535673c2c1ab41c95835d/lib/filesystem.js#L65-L80)

``` js
const transformed = options.transform && options.transform(p)
if (transformed) {
  const tmpdir = await fs.mkdtemp(path.join(os.tmpdir(), 'asar-'))
  const tmpfile = path.join(tmpdir, path.basename(p))
  const out = fs.createWriteStream(tmpfile)
  const readStream = fs.createReadStream(p)

  await pipeline(readStream, transformed, out)
  file.transformed = {
    path: tmpfile,
    stat: await fs.lstat(tmpfile)
  }
  size = file.transformed.stat.size
} else {
  size = file.stat.size
}
```

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -6,7 +6,7 @@ export type CreateOptions = {
   globOptions?: GlobOptions;
   ordering?: string;
   pattern?: string;
-  transform?: (filePath: string) => string;
+  transform?: (filePath: string) => NodeJS.ReadWriteStream | void;
   unpack?: string;
   unpackDir?: string;
 };
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -6,7 +6,7 @@ export type CreateOptions = {
   globOptions?: GlobOptions;
   ordering?: string;
   pattern?: string;
-  transform?: (filePath: string) => NodeJS.ReadWriteStream;
+  transform?: (filePath: string) => NodeJS.ReadWriteStream | void;
   unpack?: string;
   unpackDir?: string;
 };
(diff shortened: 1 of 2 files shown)


lib/index.d.ts
```diff
@@ -6,7 +6,7 @@ export type CreateOptions = {
   globOptions?: GlobOptions;
   ordering?: string;
   pattern?: string;
-  transform?: (filePath: string) => string;
+  transform?: (filePath: string) => NodeJS.ReadWriteStream | void;
```
malept: I don't think `void` can work in `stream.pipeline`?

toyobayashi: @malept If transform has no truthy return value, pipeline won't be called.

malept: Oh right. But I don't think we want to encourage that, I think that's mostly a check for JavaScript since there's no types.

toyobayashi: @malept and i see readme

> You can pass in a transform option, that is a function, which either returns nothing, or a stream.Transform. The latter will be used on files that will be in the .asar file to transform them (e.g. compress).

so i think `void` is good.

malept: I think `void` needs to be deprecated and then removed in the next major release.

toyobayashi: i see. i have changed it.

malept:
For some reason CI isn't working on this PR.

toyobayashi:
anything i can help?

lib/index.d.ts
malept: I thought about it some more and it makes a certain amount of sense that `void` is a valid return value. I've updated the definition & the test accordingly.

PR #201, opened 2020-07-26
add param to 'fs.unlinkSync' catch block

eraychumak:


malept:
This module currently only supports Node >= 10. If you need to make it work with an older version or some JavaScript engine that doesn't support optional catch bindings, you'll need to transpile it through something like [Babel](https://babeljs.io/docs/en/babel-plugin-proposal-optional-catch-binding).

Issue #167, opened 2019-02-26
asar.extractFile is Sync


thiagorova:
So... I just lost a couple minutes with this and thought it was probably a minor accident or something.

asar.extractFile is not returning a promise or using a callback.
I just checked out the imprint of the code and it seems you are calling disk.readFileSync. 
```
(archive, filename) {
  const filesystem = disk.readFilesystemSync(archive)
  return disk.readFileSync(filesystem, filename, filesystem.getFile(filename))
}
```

is there a way to call extractFile Asynchronously? 
Thanks

malept:
You can't with the current API, certainly. Given the [ASAR format](https://github.com/electron/asar#format), how feasible is it to even implement?

MarshallOfSound:
@malept You'd have to `pipe` the file as a stream and read the file from the correct offset as your stream encounters it.  Super hard to implement correctly and probably not worth it in the big scheme of things

thiagorova:
Humm... i see...
could you add this to the docs, then? The javascript API is not really that well documented, and that can lead to some confusion as to how it should be used. It's not really that big a deal for now anyways. Thanks!

malept:
We'd be happy to review a pull request to update the API docs for `asar.extractFile`.

thiagorova:
Cool! I'll look into that when I have some free time next week.

Slapbox:
@MarshallOfSound's explanation makes sense - but that leaves me wondering how Electron's `fs` can read individual files out of asar files?

Unless I'm crazy, I think I've used `fs.copyFileSync` in Electron to copy files out of the asar.

MarshallOfSound:
Closing per my comments above

PR #209, opened 2020-11-29
feat: Extract All Async, Extract File Async

waseemshahwan:
New to open source contributions. All criticism is welcome.

I've had this problem myself when using with an electron process; the window shows as non-responding when extracting which I would have liked to fix. I also found that someone had this problem in issue #167  and it's been open for a while.

### DIFF
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -4,10 +4,9 @@ const fs = require('./wrapped-fs')
 const path = require('path')
 const minimatch = require('minimatch')
 
-const { promisify } = require('util');
-const fsSymlink = promisify(fs.symlink);
-const fsUnlink = promisify(fs.unlink);
-const fsWriteFile = promisify(fs.writeFile);
+const { promisify } = require('util')
+const fsSymlink = promisify(fs.symlink)
+const fsUnlink = promisify(fs.unlink)
 
 const Filesystem = require('./filesystem')
 const disk = require('./disk')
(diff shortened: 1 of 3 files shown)


malept:
How does this solution address https://github.com/electron/asar/issues/167#issuecomment-467595078 ?

waseemshahwan:
> How does this solution address [#167 (comment)](https://github.com/electron/asar/issues/167#issuecomment-467595078) ?

Not sure what is meant in that comment. I just substituted all filesystem operation functions to use their async counterparts (fs.readFile vs fs.readFileSync). This does not block the event loop and therefore can be wrapped in a promise and run asynchronously. 

After some tests, this method worked. The process was no longer being blocked by these operations and the output was as expected.

malept:
I believe it means that there's an inherent flaw in the file format where in certain cases, it's not safe to be asynchronously read the archive, especially with `extractAll`.

waseemshahwan:
> I believe it means that there's an inherent flaw in the file format where in certain cases, it's not safe to be asynchronously read the archive, especially with extractAll.

Unless there is a fundamental difference between the native fs sync and non-sync functions, I can't imagine that to be the case. If the original commentator @MarshallOfSound can produce this potential flaw, I can look into possible fixes.

The fs.read and fs.readSync should always have the same behavior, and if they don't, then that is a flaw in the native package itself.

ibrahimelement:
Waiting for this to be approved, desperately needed right now... Our team has been tracking down a bug related to this issue... Thanks waseem!

Slapbox:
Any update on this? Looks promising.

Slapbox:
@waseemshahwan why closed? I think this would be a great addition!

waseemshahwan:
If authors of the repo are interested in merging I can reopen.

Slapbox:
@zcbenz @MarshallOfSound

PR #211, opened 2021-01-29
fix: js can precisely present MAX_SAFE_INTEGER

ChALkeR:
I'm actually unsure: why was that claim put there?
Or perhaps there is some other reason and I am missing it?

### DIFF
--- a/lib/filesystem.js
+++ b/lib/filesystem.js
@@ -6,8 +6,6 @@ const path = require('path')
 const { promisify } = require('util')
 const stream = require('stream')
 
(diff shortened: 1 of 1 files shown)


malept:
I'm going to defer to @zcbenz as to whether this is a good change. Although I am a little leery about it because I don't know if the equivalent asar support in Electron itself will work with the change.

MarshallOfSound:
This limitation is from the current ASAR implementation in Electron itself.  The `size` is limited to `int` and is cast to `uint32_t`.  Until that implementation changes the current limit is accurate

PR #221, opened 2021-08-30
feat: add file hashes to asar header

MarshallOfSound:
This also adds a new public API for `getRawHeader` to allow callers to determine the SHA of the header for integrity purposes.  This is fully backwards compatible in that no code _expects_ this field to be populated in this module as it is intended to be used by consumers rather than this module directly.


lib/index.d.ts
```diff
@@ -44,6 +44,13 @@ export type InputMetadata = {
   }
 };
 
+export type ArchiveHeader = {
+  // The JSON parsed header string
+  header: any;
```
malept, line 49: Hmm. Seems like this should be a bit more strict than `any`, since we control what the header looks like.

malept (commented):
Shouldn't this be spec'd out in the README?

lib/index.d.ts
MarshallOfSound: Fair enough, added the type info 👍

MarshallOfSound:
> Shouldn't this be spec'd out in the README?

Ah didn't realize the README actually documented this format clearly, updated the README 👍

lib/crawlfs.js
```diff
@@ -9,22 +9,32 @@ async function determineFileType (filename) {
   const stat = await fs.lstat(filename)
   if (stat.isFile()) {
     return { type: 'file', stat }
-  } else if (stat.isDirectory()) {
-    return { type: 'directory', stat }
   } else if (stat.isSymbolicLink()) {
     return { type: 'link', stat }
+  } else if (stat.isDirectory()) {
+    return { type: 'directory', stat }
```
nornagon, line 15: any particular reason for swapping these?

lib/index.d.ts
```diff
@@ -44,6 +44,27 @@ export type InputMetadata = {
...
+};
+
+export type FileRecord = {
+  offset: string;
+  size: number;
+  executable?: boolean;
+  integrity: {
+    hash: string;
+    algorithm: 'SHA256';
+  };
```
nornagon, line 58: wrong now w/ blocks format

README.md
```diff
@@ -187,6 +199,9 @@ precisely represent UINT64 in JavaScript `Number`. `size` is a JavaScript
 because file size in Node.js is represented as `Number` and it is not safe to
 convert `Number` to UINT64.
 
+`integrity` is an object consisting of a hashing `algorithm` and a hex encoded
+`hash` value.
```
nornagon, line 203: info here needs to be updated

lib/integrity.js
```diff
@@ -0,0 +1,62 @@
...
+        blocks.push(Buffer.concat(currentBlock))
+        currentBlock = []
+        callback()
+      }
+    })
+  )
+
+  return {
+    algorithm: ALGORITHM,
+    hash: hash.read(),
```
nornagon, line 56: do we use the whole-file hash for anything? it seems redundant to store the whole-file hash as well as the per-block hashes.

lib/integrity.js
```diff
@@ -0,0 +1,62 @@
...
+    fs.createReadStream(path),
+    new stream.PassThrough({
+      decodeStrings: false,
+      transform (_chunk, encoding, callback) {
+        function handleChunk (chunk) {
+          const diffToSlice = Math.min(BLOCK_SIZE - currentBlockSize, chunk.byteLength)
+          currentBlockSize += diffToSlice
+          currentBlock.push(chunk.slice(0, diffToSlice))
+          if (currentBlockSize === BLOCK_SIZE) {
+            blocks.push(Buffer.concat(currentBlock))
```
nornagon, line 35: maybe hash the block here rather than at the end to avoid reading the entire file into memory?

nornagon (changes requested):

lib/crawlfs.js
MarshallOfSound: Uh, I think I was debugging something, can be swapped back

lib/integrity.js
MarshallOfSound: We use the whole file hash when reading the entire file in `CopyFileOut` and `fs.read`.  Makes the code substantially simpler

README.md
```diff
@@ -153,12 +153,20 @@ Structure of `header` is something like this:
                "ls": {
                  "offset": "0",
                  "size": 100,
-                 "executable": true
+                 "executable": true,
+                 "integrity": {
+                   "algorithm": "SHA256",
+                   "hash": "..."
```
nornagon, line 159: needs to be updated

lib/asar.js
```diff
@@ -199,6 +203,9 @@ module.exports.extractAll = function (archive, dest) {
       // it's a file, extract it
       const content = disk.readFileSync(filesystem, filename, file)
       fs.writeFileSync(destFilename, content)
+      if (file.executable) {
+        fs.chmodSync(destFilename, '755')
+      }
```
nornagon, line 208: this seems unrelated to the rest of the PR?

lib/crawlfs.js
```diff
@@ -20,11 +20,21 @@ module.exports = async function (dir, options) {
...
+  const links = []
   const filenames = results.map(([filename, type]) => {
     if (type) {
       metadata[filename] = type
+      if (type.type === 'link') links.push(filename)
     }
     return filename
+  }).filter((filename) => {
+    // Newer glob can return files inside symlinked directories, to avoid
+    // those appearing in archives we need to manually exclude theme here
```
nornagon, line 32: can we turn this off in the glob options? looks like `{follow: false}` might do it.

lib/integrity.js
```diff
@@ -0,0 +1,66 @@
...
+          if (currentBlockSize === BLOCK_SIZE) {
+            blocks.push(hashBlock(Buffer.concat(currentBlock)))
+            currentBlock = []
+            currentBlockSize = 0
+          }
+          if (diffToSlice < chunk.byteLength) {
+            handleChunk(chunk.slice(diffToSlice))
+          }
+        }
+        handleChunk(_chunk)
```
nornagon, line 47: theoretically we could produce the block hashes and the whole-file hash in a single pass through the file, which would be faster as we wouldn't have to read the file twice. probably unnecessary though, this is likely fast enough already.

lib/crawlfs.js
MarshallOfSound: follow: false is the default, and I tried that initially 🤷

lib/asar.js
MarshallOfSound: Yeah, was just a bug that happened to be fixed as part of running the tests 🤷

PR #222, opened 2021-09-19
feat: add flags to ignore files/directories

1bitphoenix:
Add '--ignore-dir' flag to not include directories and '--ignore' flag
to not include files while asar packaging

Signed-off-by: Rituka Patwal <ritukapatwal21@gmail.com>


1bitphoenix:
@erickzhao @malept 
If anyone could review this PR? Not sure who should I tag here.

ckerr (dismissed):
The idea and code both look fine, thank you for this patch!

However it doesn't look like you added anything to test/ to exercise these code paths. Would it be possible to add some tests for this feature?

malept:
Generally speaking, instead of adding a new pair of options, I'd prefer to have `unpack` and `unpackDir` be enhanced to take an array of globs in addition to a single glob. The `glob` module should be able to specify ignores (just like the `files` array in `package.json`).

1bitphoenix:
H @ckerr,
I have update the PR by adding some tests. PTAL.

(PS: Sorry for the delay.)

1bitphoenix:
> The idea and code both look fine, thank you for this patch!
> 
> However it doesn't look like you added anything to test/ to exercise these code paths. Would it be possible to add some tests for this feature?

Hi @ckerr,
I have update the PR by adding some tests. PTAL.

(PS: Sorry for the delay.)

1bitphoenix:
> Generally speaking, instead of adding a new pair of options, I'd prefer to have `unpack` and `unpackDir` be enhanced to take an array of globs in addition to a single glob. The `glob` module should be able to specify ignores (just like the `files` array in `package.json`).

@malept 
I'm a little lost on how can we figure out from the list of globs which glob user wishes to unpack and which glob user wishes to ignore. Can you please explain your thoughts in detail?

1bitphoenix:
@ckerr 
Just a friendly ping. Can you review the PR with tests added?

PR #227, opened 2022-01-24
Add error message for file not found

Edythator:
Add an error message when trying to locate a non-existing file inside an archive when using `extract-file`.


substanc3-dev:
Tests seem to be failing, have you looked into what might be the cause of that?

erickzhao:
Seems like we need to update the xcode version for macOS CI to boot up. ref #223

malept (changes requested):
I'm OK with the spirit of the PR but not the implementation.

- `asar` is first and foremost a library, not a CLI. As such, it cannot `process.exit` in the middle.
- this needs tests so that any future changes don't regress the change.

Edythator:
Hello, @malept.

Sorry for not having understood the project's layout. I've now updated the branch in order to more accurately fit the requirements that this project needs. Please let me know if you have any further concerns.

lib/filesystem.js
```diff
@@ -139,6 +139,10 @@ class Filesystem {
     followLinks = typeof followLinks === 'undefined' ? true : followLinks
     const info = this.getNode(p)
 
+    if (typeof info === 'undefined') {
+      return undefined
```
malept, line 143: Why not throw an error here?

bin/asar.js
```diff
@@ -61,8 +61,12 @@ program.command('extract-file <archive> <filename>')
   .alias('ef')
   .description('extract one file from archive')
   .action(function (archive, filename) {
-    require('fs').writeFileSync(require('path').basename(filename),
-      asar.extractFile(archive, filename))
+    const fileData = asar.extractFile(archive, filename)
+    if (typeof fileData === 'undefined') {
+      console.error(`The file "${filename}" was not found in the archive.`)
+      return
```
malept, line 67: This is in the CLI part of the repository, so it makes more sense here to `process.exit(1)`... although if an exception is thrown, the changes here probably aren't needed.

malept (changes requested):

bin/asar.js
Edythator: Yeah, I couldn't really add a `process.exit(1)` to start with since that messes up the test.

lib/filesystem.js
Edythator: Fixed in c555f9cf73dad857f9f2974d8cadcb6ecafb5745.

PR #241, opened 2022-07-29
feat: give better hints when failing due to bad links

ckerr:
Small tweak to improve the "file links out of the package" error message.

Inspired by https://github.com/electron/electron-rebuild/issues/1024. Giving more information in this error message could make it easier to understand what the problem is and how to solve it.


VerteDinde (approved):

erickzhao (approved):

malept:
Seems more like a fix than a feat, but 🤷‍♂️

PR #244, opened 2022-10-21
Don't give up when unable to extract some files

GraemeARobinson:
When asar is extracting all files in an archive, it gives up when it encounters an error extracting one of them. This means that there are some files that can be extracted, but aren't. This change adds a try/catch block around file extraction so that the rest of the files are extracted if there is an issue with only one or a subset of the files in the archive.

### DIFF
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -210,11 +210,10 @@ module.exports.extractAll = function (archive, dest) {
         }
       } catch (e) {
         extractionErrors.push(e)
-        console.log('Unable to extract' + filename);
       }
     }
   }
-  if (extractionErrors) {
+  if (extractionErrors.length) {
     throw AggregateError(extractionErrors, 'Unable to extract some files')
   }
 }
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -208,7 +208,7 @@ module.exports.extractAll = function (archive, dest) {
           fs.chmodSync(destFilename, '755')
         }
       } catch {
-        console.log('Unable to extract' + filename);
+        console.log('Unable to extract' + filename)
       }
     }
   }
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -214,7 +214,9 @@ module.exports.extractAll = function (archive, dest) {
     }
   }
   if (extractionErrors.length) {
-    throw AggregateError(extractionErrors, 'Unable to extract some files')
+    throw new Error(
+      'Unable to extract some files:\n\n' +
+      extractionErrors.map(error => error.stack).join('\n\n'))
   }
 }
 
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -206,9 +206,9 @@ module.exports.extractAll = function (archive, dest) {
         fs.writeFileSync(destFilename, content)
         if (file.executable) {
           fs.chmodSync(destFilename, '755')
-        } catch {
-          console.log('Unable to extract' + filename);
         }
+      } catch {
+        console.log('Unable to extract' + filename);
       }
     }
   }
--- a/lib/asar.js
+++ b/lib/asar.js
@@ -200,11 +200,15 @@ module.exports.extractAll = function (archive, dest) {
       const linkTo = path.join(relativePath, path.basename(file.link))
       fs.symlinkSync(linkTo, destFilename)
     } else {
-      // it's a file, extract it
-      const content = disk.readFileSync(filesystem, filename, file)
-      fs.writeFileSync(destFilename, content)
-      if (file.executable) {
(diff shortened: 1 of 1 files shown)


lib/asar.js
```diff
@@ -200,11 +200,15 @@ module.exports.extractAll = function (archive, dest) {
...
-      fs.writeFileSync(destFilename, content)
-      if (file.executable) {
-        fs.chmodSync(destFilename, '755')
+      // it's a file, try to extract it
+      try {
+        const content = disk.readFileSync(filesystem, filename, file)
+        fs.writeFileSync(destFilename, content)
+        if (file.executable) {
+          fs.chmodSync(destFilename, '755')
+        } catch {
```
mmaietta, line 209: You're missing a `}` here. The `catch` is currently on the `if` statement

GraemeARobinson: You're right, sorry about that. I actually tested the code this time but it looks like it's still failing the automated tests and I can't see why.

MarshallOfSound (changes requested):
An error should still be thrown, just logging and continuing is incorrect behavior.

Maybe collect all caught errors and throw an Aggregate Error at the end

GraemeARobinson:
> An error should still be thrown, just logging and continuing is incorrect behavior.
Fair point.
> Maybe collect all caught errors and throw an Aggregate Error at the end
I've tried this but it's failing the lint now. Does the version of Node that the linter runs on support this feature?

I'm not a JS dev and I'm starting to get a bit out of my depth. The concept of aggregate errors is new to me so I'm not even sure if I've handled it in a logical way. 😬

lib/asar.js
```diff
@@ -200,14 +201,22 @@ module.exports.extractAll = function (archive, dest) {
...
+      // it's a file, try to extract it
+      try {
+        const content = disk.readFileSync(filesystem, filename, file)
+        fs.writeFileSync(destFilename, content)
+        if (file.executable) {
+          fs.chmodSync(destFilename, '755')
+        }
+      } catch (e) {
+        extractionErrors.push(e)
+        console.log('Unable to extract' + filename);
```
MarshallOfSound, line 213: We shouldn't be logging still

lib/asar.js
```diff
@@ -200,14 +201,22 @@ module.exports.extractAll = function (archive, dest) {
...
       }
     }
   }
+  if (extractionErrors) {
```
MarshallOfSound, line 217:
```suggestion
  if (extractionErrors.length) {
```

GraemeARobinson: 👍🏻

lib/asar.js
GraemeARobinson: 👍🏻

GraemeARobinson:
@MarshallOfSound, I'm unsure what changes are required here. Are you able to clarify please?

erikian:
@GraemeARobinson sorry about intruding your PR, but it looks like `AggregateError` is only available in Node.js >= 15, and this package should support Node.js >= 10 as per the `engines` field in `package.json`. I believe you can use something like this to display all the errors in a more compatible way:


```js
if (extractionErrors.length) {
  extractionErrors.forEach(error => console.error(error))
  throw new Error(`Unable to extract some files:\n\n ${extractionErrors.map(error => error.message).join('\n\n')}`)
}
```

This should also make the linter happy as it doesn't seem to recognize `AggregateError`.

erikian:
The following code should show the error message _and_ the error stack (while `error.stack` is non-standard, it seems to be the only option for displaying the full error here):

```js
if (extractionErrors.length) {
  throw new Error(`Unable to extract some files:\n\n ${extractionErrors.map(error => error.stack).join('\n\n')}`)
}
```

GraemeARobinson:
> The following code should show the error message _and_ the error stack (while `error.stack` is non-standard, it seems to be the only option for displaying the full error here):
> 
> ```js
> if (extractionErrors.length) {
>   throw new Error(`Unable to extract some files:\n\n ${extractionErrors.map(error => error.stack).join('\n\n')}`)
> }
> ```Thank you @erikian, all checks passed with your suggestion :-)

BlackHole1 (approved):

GraemeARobinson:
@MarshallOfSound, are you happy with the changes? Your request for change is blocking this merge and I think I've addressed the issue you pointed out

GraemeARobinson:
@BlackHole1 , what else needs to be done for this to be merged?

BlackHole1:
> @BlackHole1 , what else needs to be done for this to be merged?

From my own subjective viewpoint, it seems there is no further revision necessary for this PR, yet @MarshallOfSound's verification is still required.

erikian (approved):

Issue #144, opened 2018-03-16
Failing with scoped node_modules


davej:
Asar seems to fail with modules that are scoped, e.g.. `@scope/project-name`.

Getting the following error while asar is iterating over the `node_modules/@pnpm/default-fetcher` directory:

```
/Users/dave/code/my-app/app/node_modules/asar/lib/filesystem.js:20
        json = json.files[dir]
                    ^

TypeError: Cannot read property 'files' of undefined
    at Filesystem.searchNodeFromDirectory (/Users/dave/code/my-app/app/node_modules/asar/lib/filesystem.js:20:21)
    at Filesystem.searchNodeFromPath (/Users/dave/code/my-app/app/node_modules/asar/lib/filesystem.js:30:23)
    at Filesystem.insertDirectory (/Users/dave/code/my-app/app/node_modules/asar/lib/filesystem.js:41:23)
    at handleFile (/Users/dave/code/my-app/app/node_modules/asar/lib/asar.js:120:20)
    at next (/Users/dave/code/my-app/app/node_modules/asar/lib/asar.js:160:12)
    at /Users/dave/code/my-app/app/node_modules/asar/lib/asar.js:161:14
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

fczuardi:
I am not sure if this is the same issue, but I had a similar error message when trying it with the pattern

```
'/{electron.js,package.json,dist/**/*,node_modules/electron-log/**/*}'
```
My workaround after debugging this filesystem.js file was to change my pattern to include the parent dir before the subdir, like this:

```
'/{electron.js,package.json,dist/**/*,node_modules,node_modules/electron-log/**/*}'
```

x-yuri:
I ran into the error when extracting a file this way:

```
$ asar ef ~/.cache/Cypress/3.4.1/Cypress/resources/electron.asar /browser/api/web-contents.js
```

In my case the fix was to strip the leading slash:

```
$ asar ef ~/.cache/Cypress/3.4.1/Cypress/resources/electron.asar browser/api/web-contents.js
```

Otherwise the first item in the list is [empty][1].

[1]: https://github.com/electron/asar/blob/v2.0.2/lib/filesystem.js#L20

RobinCodeX:
My asar:
```
\index.js
\package.json
\preload.js
\dist
\dist\index.html
\dist\assets
\dist\assets\index.810da0b0.css
[...]

PR #246, opened 2022-10-27
fix: allowing nested empty directories. Fixes #245 and #144

### LINKED ISSUE #245: Create asar fails for nested empty directories: `Cannot read property 'files' of undefined`
Version: 3.2.0

This can be easily reproduced by having a folder structure below:
- out
- - scripts
- - - main.js
- - - preload.js
- - - renderer.js

This is notably a big issue as it also causes #144, where scoped modules are often a "double-stack" of folders, the scope being an empty root folder for the `package` subfolder

I'm using `asar` programmatically. With the above folder structure, I'm simply doing:
```
const options: CreateOptions = {
      pattern: `/out/**/*`
}
await createPackageWithOptions(path.resolve(this.src), this.outFile, options)
```

For my project, I do have the ability to change the file/folder structure. But we're fully blocked when it comes to scoped node_modules
Stacktrace:
```
  ⨯ TypeError: Cannot read property 'files' of undefined
    at Filesystem.searchNodeFromPath (/eb-test/node_modules/asar/lib/filesystem.js:37:14)
    at Filesystem.insertDirectory (/eb-test/node_modules/asar/lib/filesystem.js:47:23)
    at handleFile (/eb-test/node_modules/asar/lib/asar.js:118:20)
    at next (/eb-test/node_modules/asar/lib/asar.js:148:11)
    at Object.module.exports.createPackageFromFiles (/eb-test/node_modules/asar/lib/asar.js:152:10)
    at Object.module.exports.createPackageWithOptions (/eb-test/node_modules/asar/lib/asar.js:40:25)
```

mmaietta:
Allowing nested empty directories, such as for the case with scoped node_modules.
Fixes #245 and #144

I tried to think of an additional test to add but I think this one was already supposed to cover this use case? https://github.com/electron/asar/tree/main/test/input/packthis-unicode-path/dir1
https://github.com/electron/asar/blob/main/test/api-spec.js#L84-L90

### DIFF
--- a/lib/filesystem.js
+++ b/lib/filesystem.js
@@ -23,6 +23,9 @@ class Filesystem {
     const dirs = p.split(path.sep)
     for (const dir of dirs) {
       if (dir !== '.') {
+        if (!json.files[dir]) {
+          json.files[dir] = { files: {} }
+        }
         json = json.files[dir]
       }
     }
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   "devDependencies": {
     "@continuous-auth/semantic-release-npm": "^2.0.0",
     "@semantic-release/changelog": "^5.0.0",
-    "electron": "^15.0.0",
+    "electron": "^5.0.0",
     "electron-mocha": "^8.2.1",
     "lodash": "^4.17.15",
     "mocha": "^7.1.1",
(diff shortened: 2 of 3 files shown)


MarshallOfSound (approved):

VerteDinde (approved):

VerteDinde:
@mmaietta Thanks for the PR! 🙇‍♀️ We'll get this merged and a new release cut for you

PR #252, opened 2022-12-06
fix: bad output on unknown CLI command

CanadaHonk:
Looks like this without PR / before:
image

And with PR, performs as expected:
image


BlackHole1 (approved):
LGTM

dsanders11 (approved):

PR #253, opened 2022-12-06
fix: ensure that files/folders whose names are properties of Object.prototype are packaged/extracted correctly

### LINKED ISSUE #235: Folders named constructor not being packed properly
If a library includes a `/constructor` folder the folder is not packed into the asar. 

See:  https://github.com/trufflesuite/truffle/tree/master/packages/hdwallet-provider/src/constructor

erikian:
As noted in #235, folders (and also files) named `constructor` were not being packaged. This happened because the objects holding the directory tree were being created with `{}`, which inherits all properties from `Object.prototype`, such as `constructor` and `__proto__`, so when trying to add a folder named `constructor`, it was assumed it already existed. Now we're creating these objects without a prototype with `Object.create(null)`, so there shouldn't be any more conflicts.

Fixes #235.


erikian:
Pretty sure the test failure is git messing with me, I'll check it out when I get a chance

erikian:
Turns out it was an issue with the trailing line endings my editor was adding, looks like Windows on CI is creating a file with line endings which are different from the ones in my local Windows? Anyway, the PR is ready to review now.

BlackHole1 (approved):
LGTM. /cc @electron/wg-ecosystem

dsanders11 (approved):

PR #254, opened 2022-12-06
fix: properly handle rejections on `createPackageWithOptions`

erikian:
Minor cleanup: `asar.createPackageWithOptions` doesn't take a fourth argument, so it's safe to remove it

### DIFF
--- a/bin/asar.js
+++ b/bin/asar.js
@@ -35,11 +35,9 @@ program.command('pack <dir> <output>')
       builddir: options.sb,
       dot: !options.excludeHidden
     }
-    asar.createPackageWithOptions(dir, output, options, function (error) {
-      if (error) {
-        console.error(error.stack)
-        process.exit(1)
-      }
+    asar.createPackageWithOptions(dir, output, options).catch(error => {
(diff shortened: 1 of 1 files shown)


malept (commented):
Shouldn't the promise be handled?

erikian:
Probably, but unless we output a more useful error message or do something else on rejections, moving the callback we had to a `.catch` will do basically the same an unhandled rejection already does: output the error message, the stack and exit with a non-zero code.

malept:
I think Node.js also outputs `UnhandledRejection`, which isn't ideal.

erikian:
Good point, I was testing in node 16 which doesn't show that message anymore :sweat_smile:

I've changed `error.stack` to just `error` as `error.stack` [seems non-standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack). They both show the error message and stack for me (node 16 on Windows), and as a bonus, the `error` output has some formatting now.

erickzhao (approved):

PR #257, opened 2023-01-04
build(deps): upgrade minimatch to ^3.0.5

Bharat-dev:
Upgrade minimatch to ^3.0.5 to address security vulnerabilities

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "chromium-pickle-js": "^0.2.0",
     "commander": "^5.0.0",
     "glob": "^7.1.6",
-    "minimatch": "^3.0.4"
+    "minimatch": "^3.0.5"
   },
   "optionalDependencies": {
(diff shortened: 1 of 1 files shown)


IoneMurray1:
Hello. Is there any reason this is not being approved by code owners?

erickzhao:
Dunno if this is strictly necessary because the lockfile already specifies `3.1.2` for the `minimatch@^3.0.4` dependency.

https://github.com/electron/asar/blob/main/yarn.lock#L2698

erikian:
It might be necessary because we don't publish the lockfile to npm, only the `package.json`, so if a project already has `minimatch@3.0.4` installed, `asar` will be happy with that version.

MarshallOfSound:
This definitely isn't necessary, projects downstream can bump this dependency without us releasing entire new packages

PR #262, opened 2023-01-17
build: add Dependabot configuration

dsanders11:
Let's use Dependabot here.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
erickzhao, line 1: Does this enable dependabot for all deps? I would prefer to have this be security-only.

dsanders11: I've updated the config so that it'll be security-only, [according to this documentation](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates):

> If you only require security updates and want to exclude version updates, you can set open-pull-requests-limit to 0 in order to prevent version updates for a given package-ecosystem. For more information, see "[open-pull-requests-limit](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit)."

MarshallOfSound (changes requested):
Dependabot is already on for this repo for security updates.  You don't need a config for that

image

PR #268, opened 2023-04-09
fix: inline @types/glob, remove from deps

isaacs:
Inline types for old glob version, as discussed in https://github.com/electron/asar/pull/267


MarshallOfSound (approved):
Seems legit

PR #284, opened 2023-09-26
fix: Don't overwrite existing folders

felixrieseberg:
It's amazing that I'm the first one to hit this bug - but if you have a folder structure where `asar` just happens to visit the directory _after_ it's added files, it will overwrite the existing files. That's bad, so this is fixing that.


VerteDinde (approved):
Oof, sorry you ran into this!

PR #300, opened 2024-01-31
fix: fix bug in filesystem.insertLink

jrainlau:
In MacOS, if I run `asar.createPackage()` to pack a folder into `.asar`, which contains "symlink", it would get wrong after I run `asar.extractAll()` from the generated `.asar`.

Let's look at this Beyond Compare result:

企业微信截图_0ef755c7-3d85-4037-8d72-906e533b1fe1

The folder in left is the origin one, and the folder in right was generated by using `asar.extractAll()` from the origin one's `.asar` ( folderLeft --> folderLeft.asar --> folderRight).

Inside the left folder, file **Resources** were pointed to `Versions/Current/Resources`, but the same file inside the right folder were pointed to `Versions/A/Resources`, it's wrong.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   cfa: continuousauth/npm@2.1.0
-  node: electronjs/node@2.1.0
+  node: electronjs/node@2.2.0
 
 workflows:
   test_and_release:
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@
 .node-version
 npm-debug.log
 .idea
+.vscode
```
ckerr, line 11: This change seems to be unrelated to the fix

jrainlau: Sorry for my mistake. I just removed it in 07a14a8

PR #304, opened 2024-03-07
chore: bump glob to 9.3.5

theogravity:
This updates the project to use `glob@9.3.5` (not using glob@10 as that looks to be ESM-only and this project does not have ESM support).

The reason for updating `glob` is older versions of `glob` uses `inflight`, which has a [medium-level vulun](https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116) that triggers our Vanta SOC2 compliance monitoring for security issues.

This version of glob no longer works for node 14, so I bumped the engine requirements to 18.

### DIFF
--- a/lib/crawlfs.js
+++ b/lib/crawlfs.js
@@ -1,9 +1,7 @@
 'use strict'
 
-const { promisify } = require('util')
-
 const fs = require('./wrapped-fs')
-const glob = promisify(require('glob'))
+const { glob } = require('glob')
 
 async function determineFileType (filename) {
   const stat = await fs.lstat(filename)
@@ -18,8 +16,13 @@ async function determineFileType (filename) {
 
 module.exports = async function (dir, options) {
   const metadata = {}
-  const crawled = await glob(dir, options)
+  let crawled = await glob(dir, options || {})
+
+  // https://github.com/isaacs/node-glob/issues/576
+  crawled = crawled.sort((a, b) => a.localeCompare(b, 'en'))
+
   const results = await Promise.all(crawled.map(async filename => [filename, await determineFileType(filename)]))
+
   const links = []
   const filenames = results.map(([filename, type]) => {
     if (type) {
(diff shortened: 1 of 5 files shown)


lib/crawlfs.js
```diff
@@ -18,8 +16,13 @@ async function determineFileType (filename) {
 
 module.exports = async function (dir, options) {
   const metadata = {}
-  const crawled = await glob(dir, options)
+  let crawled = await glob(dir, options || {})
```
theogravity, line 19: I did try `async function (dir, options = {})` but tests would fail with that.

dsanders11 (changes requested):
This needs to be marked as a breaking change (add `BREAKING CHANGE: ...` in the body of the commit message) so that it would trigger a v4 release on merge.

If we're going to do a breaking change we probably want to just jump to Node v18 rather than v16 which is already EOL.

But I'm not sure if we consider this worth introducing a breaking change to address.

dsanders11:
Tests appear to be failing on Windows.

theogravity:
@dsanders11 Addressed the comments.

erickzhao (changes requested):
I'm not sure if this is a complete solution for the purposes of resolving security scannner warnings because we still have dependencies to `glob@7.2.0, glob@^7.0.5, glob@^7.1.3` as per the `yarn.lock` file via various transient dependencies.

theogravity:
> I'm not sure if this is a complete solution for the purposes of resolving security scanner warnings because we still have dependencies to `glob@7.2.0, glob@^7.0.5, glob@^7.1.3` as per the `yarn.lock` file via various transient dependencies.

The transients are only on devDeps though. At least for our scanner, it only cares about deps for node_module installs.

ckerr (approved):
The code changes LGTM, approving on that part.

@dsanders11 is right about the node bump / breaking change question. I don't have a strong opinion there except that we probably shouldn't depend on EOL packages  

@erickzhao is right that there are still other devdep vectors do depend on old versions of glob, but fixing one is better than fixing zero :man_shrugging:

hallelk:
bump

PR #318, opened 2024-08-12
Enable `getNode` to follow directory links (fixes #248)

### LINKED ISSUE #248: There is an error that seems to be caused by soft links
### wrong content：
image

### temporary solution：
- Modify this line of code for fault tolerance
image

AndreasFranek:
Fixes #248 

As reported in the issue, soft links (or symlinks/symbolic links) can cause a `TypeError` to occur in `asar.stat`. This happens when a directory below the stat'ed path is actually a link.

This commit adds a test demonstrating the problem and fixes the problem by introducing a `followLinks` variable in `getNode` and resolving directory links, if it is set.

This is my first PR in this project so any feedback is welcome. I was not sure whether to implement the fix imperatively or recursively, for example. I.e., it is possible to follow the link like this instead:

```JS
  getNode (p, followLinks) {
    followLinks = typeof followLinks === 'undefined' ? true : followLinks
    var node = this.searchNodeFromDirectory(path.dirname(p))
    while (node.link && followLinks) {
      node = this.searchNodeFromDirectory(node.link)
    }
    const name = path.basename(p)
    if (name) {
      return node.files[name]
    } else {
      return node
    }
  }
```


AndreasFranek:
Follow-up question, why is the `followLinks` parameter not defaulted to `true`?
I.e., why is it 
```JS
  getFile (p, followLinks) {
    followLinks = typeof followLinks === 'undefined' ? true : followLinks
```
and not 
```JS
  getFile (p, followLinks = true) {
```
?

MarshallOfSound:
Should probably land after #320

AndreasFranek:
Rebased.

AndreasFranek:
Any further input on this?

test/api-spec.js
```diff
@@ -168,4 +168,8 @@ describe('api', function () {
 
     assert.deepStrictEqual(topLevelFunctions, defaultExportFunctions);
   });
+  it('should stat a symlinked file', async () => {
+    const stats = asar.statFile('test/input/stat-symlink.asar', 'real.txt', true);
+    return assert.strictEqual(stats.size, 19);
```
erikian, line 173: Can we assert that this is not a link instead of checking the size? I think it makes the intent clearer:

```suggestion
    return assert.strictEqual(stats.link, undefined);
```

AndreasFranek: Good point. Done.

erikian (approved):

PR #320, opened 2024-08-16
refactor: rewrite to typescript retaining API

### LINKED ISSUE #251: Typescript rewrite?
Any plans to rewrite this library in Typescript? Would a PR be welcome?

MarshallOfSound:
Modernize the codebase substantially. Key things:
* Git claims I rewrote these files from scratch, I did not, the actual changes are _just_ adding types and in a few edges cases adding additional error handling (that typescript revealed as possible)
* I did not touch the tests at all, I had to fix like 5 imports, and prettier added a bunch of semicolons, but the actual tests were not "fixed" or "changed" at all. They will be converted to typescript _after_ this PR lands to ensure confidence

Closes #251


package.json
```diff
@@ -25,36 +24,31 @@
     "url": "https://github.com/electron/asar/issues"
   },
   "scripts": {
+    "build": "tsc",
```
erikian, line 27:
```suggestion
    "build": "rimraf ./lib && tsc",
```

package.json
```diff
@@ -25,36 +24,31 @@
...
-    ]
-  },
-  "tsd": {
-    "directory": "test"
+    "test": "yarn lint && yarn mocha",
+    "lint": "yarn prettier:check",
+    "prettier": "prettier \"src/**/*.ts\" \"test/**/*.ts\" \"test/**/*.js\"",
+    "prettier:check": "yarn prettier --check",
+    "prettier:write": "yarn prettier --write",
+    "prepare": "tsc"
```
erikian, line 34:
```suggestion
    "prepare": "rimraf ./lib && tsc"
```

yarn.lock
erikian, line 1: issue: `yarn.lock` gets updated for me when running `yarn`, looks like it still contains `standard` and its dependencies

erikian (commented):
just one issue regarding `yarn.lock`, otherwise looks good

src/asar.ts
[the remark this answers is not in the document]
MarshallOfSound: We shouldn't, asar only did named exports

package.json
MarshallOfSound: I don't think this is correct, it would break incremental tsc

erikian: Fair point - we should enable incremental compilation though, it's off by default

src/filesystem.ts
```diff
@@ -0,0 +1,214 @@
...
+  link: string;
+};
+
+export type FilesystemEntry = FilesystemDirectoryEntry | FilesystemFileEntry | FilesystemLinkEntry;
+
+export class Filesystem {
+  private src: string;
+  private header: FilesystemEntry;
+  private headerSize: number;
+  private offset: bigint;
```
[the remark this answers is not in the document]
MarshallOfSound: The perf overhead of using true privates is not worth it IMO, we don't have an actual need to not expose these. The semantic indicator is IMO enough

src/asar.ts
```diff
@@ -0,0 +1,265 @@
...
+import { crawl as crawlFilesystem, determineFileType } from './crawlfs';
+import { IOptions } from 'glob';
+
+/**
+ * Whether a directory should be excluded from packing due to the `--unpack-dir" option.
+ *
+ * @param {string} dirPath - directory path to check
+ * @param {string} pattern - literal prefix [for backward compatibility] or glob pattern
+ * @param {array} unpackDirs - Array of directory paths previously marked as unpacked
+ */
```
dsanders11, line 16:
```suggestion
/**
 * Whether a directory should be excluded from packing due to the `--unpack-dir" option.
 *
 * @param dirPath - directory path to check
 * @param pattern - literal prefix [for backward compatibility] or glob pattern
 * @param unpackDirs - Array of directory paths previously marked as unpacked
 */
```

We should use TSDoc comments rather than JSDoc style after the conversion. The parameter types are redundant and often go stale.

src/asar.ts
```diff
@@ -0,0 +1,265 @@
...
+};
+
+export async function createPackageWithOptions(src: string, dest: string, options: CreateOptions) {
+  const globOptions = options.globOptions ? options.globOptions : {};
+  globOptions.dot = options.dot === undefined ? true : options.dot;
+
+  const pattern = src + (options.pattern ? options.pattern : '/**/*');
+
+  const [filenames, metadata] = await crawlFilesystem(pattern, globOptions);
+  return module.exports.createPackageFromFiles(src, dest, filenames, metadata, options);
```
dsanders11, line 49:
```suggestion
  return createPackageFromFiles(src, dest, filenames, metadata, options);
```

src/asar.ts
```diff
@@ -0,0 +1,265 @@
...
+    if (!metadata[filename]) {
+      const fileType = await determineFileType(filename);
+      if (!fileType) {
+        throw new Error('Unknown file type for file: ' + filename);
+      }
+      metadata[filename] = fileType;
+    }
+    const file = metadata[filename];
+
+    let shouldUnpack;
```
dsanders11, line 135:
```suggestion
    let shouldUnpack: boolean;
```

src/disk.ts
```diff
@@ -0,0 +1,150 @@
...
+    out.write(sizeBuf);
+    return out.write(headerBuf, () => resolve());
+  });
+  return writeFileListToStream(dest, filesystem, out, fileList, metadata);
+}
+
+export function readArchiveHeaderSync(archivePath: string) {
+  const fd = fs.openSync(archivePath, 'r');
+  let size;
+  let headerBuf;
```
dsanders11, line 86:
```suggestion
  let size: number;
  let headerBuf: Buffer;
```

src/filesystem.ts
```diff
@@ -0,0 +1,214 @@
...
+    const dirNode = this.searchNodeFromPath(path.dirname(p)) as FilesystemDirectoryEntry;
+    const node = this.searchNodeFromPath(p) as FilesystemFileEntry;
+    if (shouldUnpack || dirNode.unpacked) {
+      node.size = file.stat.size;
+      node.unpacked = true;
+      node.integrity = await getFileIntegrity(p);
+      return Promise.resolve();
+    }
+
+    let size;
```
dsanders11, line 123:
```suggestion
    let size: number;
```

src/integrity.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const blockHashes: string[] = [];
+  let currentBlockSize = 0;
+  let currentBlock: Buffer[] = [];
+
+  await pipeline(
+    fs.createReadStream(path),
+    new stream.PassThrough({
+      decodeStrings: false,
+      transform(_chunk, encoding, callback) {
```
dsanders11, line 34:
```suggestion
      transform(_chunk: Buffer, encoding, callback) {
```

dsanders11 (commented):
Not blockers, but a comment about TSDoc and some suggestions for improving type coverage.

dsanders11 (approved):

erikian (approved):

PR #322, opened 2024-09-10
fix: ensure symlinks do not write unexpectedly

MarshallOfSound:
Ensures we don't write symlinks or files outside of the destination directory as this is unexpected.


erickzhao (approved):
:shipit:

PR #327, opened 2024-10-01
fix: re-remove @types/glob from deps

jw-maynard:
The @types/glob was removed from `optionalDependencies` by the package maintainer @isaacs in #268. The glob package is now written in TS and shouldn't need @types.

I did see where he originally inlined all the types under lib/index.d.ts but I'm guessing that's no longer needed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "prepare": "tsc"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jw-maynard:
I see that this would require some major updates like bumping the minimum node version

PR #332, opened 2024-10-11
Move @types/glob to devDependencies

### LINKED ISSUE #330: Please remove dependency on `@types/glob` which is not required and causes breakage in typedoc builds
The version of `@types/glob` that is used by this package is incompatible with newer versions of glob or minimatch. Those libraries are written directly in typescript (this wasn't the case before) so no types package is necessary. Please remove `@types/glob` as a dependency.

Some additional context where this has come up: https://github.com/isaacs/rimraf/issues/264

Hinton:
Type dependencies should be dev dependencies, that way they won't affect downstream packages.

This change seems to have no negative impacts on building or running tests in the project.

Resolves #330

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,12 +34,12 @@
     "prepare": "tsc"
   },
   "dependencies": {
-    "@types/glob": "^7.1.0",
     "commander": "^5.0.0",
     "glob": "^7.1.6",
     "minimatch": "^3.0.4"
   },
   "devDependencies": {
+    "@types/glob": "^7.1.0",
     "@types/minimatch": "^3.0.5",
     "@types/node": "^12.0.0",
     "electron": "^22.0.0",


MarshallOfSound (changes requested):
This is semantically incorrect and will also cause issues as we do use glob types in exported types

aaclayton:
This is causing big problems for us too. Our documentation builds (typedoc) are completely failing because of this unless we do clumsy manual workarounds.

erikian:
> Isn't glob now natively in typescript anyways? Why is an extra types package needed?

It is, but at this moment `@electron/asar` [supports Node 10.12.0 and up](https://github.com/electron/asar/blob/b7730638891726666776c68116b51043d39424ef/package.json#L14-L16), while `glob` [only became native TS on v9](https://github.com/isaacs/node-glob/blob/08958fa38ac1a59f82b0cf912208ec9548fb28cc/changelog.md#90) which requires Node 16. [We do have plans](https://github.com/electron/governance/tree/13acc859f71d524e9e67b748741d1c9842079ce1/wg-ecosystem#nodejs-upgrade-policy) to upgrade all ecosystem packages to a newer Node version as soon as support for `require`ing ESM graphs makes it into Node LTS, but I'm afraid we probably won't be updating `glob` until then unless we need to land another breaking change in the meantime.

That being said, based on https://github.com/electron/asar/pull/267 / https://github.com/electron/asar/pull/268, looks like we've solved this in the past by inlining the types, and that got lost when `@electron/asar` was rewritten in TS in #320, so we can probably just re-land the changes from #268 – I'll open a PR for that in a minute.

erikian:
Fixed in #336

Issue #331, opened 2024-10-07
`isUnpackedDir` unintentionally unpacks other directories with same dir name prefix


mmaietta:
Hello all

I'm working on migrating electron-builder to leverage electron/asar instead of its in-house solution. In one of our unit tests I have the following structure:
```
/project# tree "/tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-jp88Jv/asar-app-1"
/tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-jp88Jv/asar-app-1
├── app
│   ├── package.json
│   │   └── readme.md
│   └── readme.md
├── index.html
├── index.js
├── node_modules
│   ├── @electron-builder
│   │   ├── test-smart-unpack
│   │   │   ├── foo.dll
│   │   │   └── package.json
│   │   └── test-smart-unpack-empty
│   │       └── package.json
```
`test-smart-unpack` is auto-detected by electron-builder to unpack due to the underlying `.dll` file, and in turn, that minimatch is passed to electron/asar to unpack that dir with the pattern below
```
pattern = "{node_modules/@electron-builder/test-smart-unpack,node_modules/edge-cs}"
```
https://github.com/electron/asar/blob/464e83436967438c74d8ac184b088fc780706b2d/src/asar.ts#L22-L31

The issue I'm running into is that `test-smart-unpack-empty` is **not** supposed to be getting unpacked as it fails the minimatch, but the `else` logic in `isUnpackedDir` is causing it to do a `startsWith` match on the folder path and then returns `true`.
![Image](https://github.com/user-attachments/assets/89bf9d4b-29d8-41a5-b3f2-7268ce9d2853)

What I think the `else` logic is attempting to do is to determine if the `dirPath` is _within_ the tree hierarchy of a previously-unpacked dir. Is that right?
https://github.com/electron/asar/blob/464e83436967438c74d8ac184b088fc780706b2d/src/asar.ts#L29
If so, I think an additional check is needed such that we check that the relative path is within the dir path hierarchy, and not just a `startsWith` check.
Maybe something like: `dirPath.startsWith(unpackDir) && !path.relative(dirPath, unpackDir).startsWith("../")`?

Wanted to post an issue here first to touch base with you all, in case I'm doing something wrong, before diving in deeper to see if I can PR something

Akarinx:
maybe it's the same problem like https://github.com/electron/asar/issues/303,  and the same fix method

mmaietta:
[...]

PR #333, opened 2024-10-15
fix: `isUnpackedDir` matches non-child directories that have same folder name prefix

mmaietta:
Fixes: https://github.com/electron/asar/issues/331 and updates test case for minimatch patterns. (Happy to write additional or separate tests if preferred)

Test project file structure:
```
/project# tree "/tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-jp88Jv/asar-app-1"
/tmp/et-db411cd21f3bbd882f3a5a30c58db6ff/t-jp88Jv/asar-app-1
├── app
│   ├── package.json
│   │   └── readme.md
│   └── readme.md
├── index.html
├── index.js
├── node_modules
│   ├── @electron-builder
│   │   ├── test-smart-unpack
│   │   │   ├── foo.dll
│   │   │   └── package.json
│   │   └── test-smart-unpack-empty
│   │       └── package.json
```
The pattern passed to `unpackDir` minimatch glob.
```
pattern = "{node_modules/@electron-builder/test-smart-unpack,node_modules/edge-cs}"
```
https://github.com/electron/asar/blob/464e83436967438c74d8ac184b088fc780706b2d/src/asar.ts#L22-L31

The issue I'm running into is that `test-smart-unpack-empty` is **not** supposed to be getting unpacked as it fails the minimatch, but the `else` logic in `isUnpackedDir` is causing it to do a `startsWith` match on the folder path and then returns `true`.
Image

This PR introduces a check to make sure that the previous folders being unpacked do not include folders that happen to have the same prefix but are not a child directory of a previously flagged unpacked directory.

Additional note, this does not change logic for non-minimatch patterns (e.g. `--unpack-dir dir2/subdir`) as that logic flow is left unchanged.
https://github.com/electron/asar/blob/464e83436967438c74d8ac184b088fc780706b2d/src/asar.ts#L23


mmaietta:
So I'm not sure why the unit tests are timing out with `Error: Timeout of 2000ms exceeded.` The timeout seems to occur on random tests between runs, not on the test case that I updated.
First timeout was on UT:
```
1) should create archive from directory (without hidden files)
```
Second timeout was on UT:
```
1) should create archive from directory with unpacked dirs and files
```

BlackHole1 (approved):

Issue #303, opened 2024-03-04
Symlinked directories not included in asar


ttpssabc:
Hi!
If a directory is pointed to by a symbolic link whose name is the prefix of the directory, the directory won't be included in the asar.
For example, the directory `aa` won't be included.
```
├── a -> aa
└── aa
```
https://github.com/electron/asar/blob/main/lib/crawlfs.js#L30

PR #335, opened 2024-10-23
fix: checking if symlink with same prefix points outside the directory

mmaietta:
> If a directory is pointed to by a symbolic link whose name is the prefix of the directory, the directory won't be included in the asar. For example, the directory `aa` won't be included.
> 
> ```
> ├── a -> aa
> └── aa
> ```

Added a check (and test case) to see if the symlink is actually to a different directory and not a file (e.g. `startsWith`) within the symlinked dir. New test case mirrors `test/input/packthis-with-symlink` but with the symlink having the same prefix of a directory outside the symlink

Fixes: https://github.com/electron/asar/issues/303


BlackHole1 (approved):
LGTM

mmaietta:
Thanks @BlackHole1 ! Would you mind also taking a look at this PR? It's the same style of fix but for asar unpacking: https://github.com/electron/asar/pull/333

BlackHole1:
@mmaietta Already approved! Thank you for your fixed!
In general, we do not merge PRs immediately (unless they are very simple or submitted by a bot for dependency updates). We usually wait a few days to allow other maintainers to provide their input.

PR #336, opened 2024-10-24
fix: inline `@types/glob`

erikian:
Basically just copy-pasted from #268; as discussed in https://github.com/electron/asar/pull/332#issuecomment-2435407933, we used to inline these types to avoid issues downstream, but that was lost when #320 landed so this is just restoring our previous workaround.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "prepare": "tsc"
   },
   "dependencies": {
-    "@types/glob": "^7.1.0",
     "commander": "^5.0.0",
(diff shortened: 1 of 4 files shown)


ckerr (approved):

BlackHole1 (approved):

t3chguy:
This doesn't appear to fix anything for me. The new type def file isn't included in the npm package. 

image

erikian:
@t3chguy good catch, I opened #339 to fix that. Thanks!

PR #339, opened 2024-10-30
fix: use local `glob` types

### LINKED ISSUE #330: Please remove dependency on `@types/glob` which is not required and causes breakage in typedoc builds

erikian:
Follow-up to #336, as pointed out in https://github.com/electron/asar/pull/336#issuecomment-2444739563 I forgot to actually replace the usages of the `IOptions` type from `glob` by our local copy.

Fixes #330.

### DIFF
--- a/src/asar.ts
+++ b/src/asar.ts
@@ -10,7 +10,7 @@ import {
 } from './filesystem';
 import * as disk from './disk';
 import { crawl as crawlFilesystem, determineFileType } from './crawlfs';
-import { IOptions } from 'glob';
+import { IOptions } from './types/glob';
 
 /**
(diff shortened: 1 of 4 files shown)


BlackHole1 (approved):

ckerr (approved):

t3chguy:
Thanks @erikian - can confirm this works: https://github.com/element-hq/element-desktop/pull/1966

PR #341, opened 2024-11-12
fix: Respect unpack minimatch for symlinks within previously unpacked directories

mmaietta:
**Summary:**
Fixes issue with unpack configuration when considering symlinks within previously unpacked directories. This directly fixes unpacking static `.framework` modules on Mac, as otherwise codesigning will fail due to symlink files/directories not being reflected in the app.asar.unpacked directory.

**Fix:**
- Tracks all symlinks in a list (similar as files) as to what are supposed to be unpacked.
- Adds `unpacked` flag as `node.unpacked=true`
- Extracts common logic for detecting if a path is unpacked
- Added promisified `symlink` and `readlink`
- Reuses Filesystem unit test approach for programmatically creating symlinks during tests so as to properly replicate symlinks on Win/Linux/MacOS. (committing Windows symlinks created with `mklink` do not get committed properly to git)

**Details:**
`.framework` has to have all symlinks reproduced in the unpacked directory for signing. The top is with electron/asar unpacked folder, the bottom is the actual framework contents.
Screenshot 2024-11-11 at 4 28 35 PM

Without this PR, codesigning will fail with `QtCore.framework: bundle format unrecognized, invalid, or unsuitable`
```
  • signing         file=dist/mac/SanitizedAppName.app platform=darwin type=distribution identity=<SanitizedIdentity> provisioningProfile=none
  ⨯ Command failed: codesign --sign <SanitizedIdentity> --force --timestamp --options runtime --entitlements ./build/entitlements.mac.plist /Users/<project-folder>/dist/mac/SanitizedAppName.app/Contents/Resources/app.asar.unpacked/node_modules/<sanitized-dep-name>/lib/Release/QtCore.framework
/Users/<project-folder>/dist/mac/SanitizedAppName.app/Contents/Resources/app.asar.unpacked/node_modules/<sanitized-dep-name>/lib/Release/QtCore.framework: bundle format unrecognized, invalid, or unsuitable
```

Fixes: https://github.com/electron-userland/electron-builder/issues/8655#issuecomment-2463084611


src/disk.ts
```diff
@@ -53,13 +56,27 @@ const writeFileListToStream = async function (
...
     }
   }
+  const unpackedSymlinks = links.filter((f) => f.unpack);
+  for (const file of unpackedSymlinks) {
+    // the symlink needs to be recreated outside in .unpacked
+    const filename = path.relative(filesystem.getRootPath(), file.filename);
+    const link = await fs.readlink(file.filename);
+    await fs.symlink(link, path.join(`${dest}.unpacked`, filename)).catch(async (error) => {
+      if (error.code === 'EPERM' && error.syscall === 'symlink') {
+        throw new Error(
```
beyondkmp, line 66: On Windows, using junction to create symlinks should avoid permission issues.

mmaietta: Great callout. I read somewhere (can't find the link atm) that junctions are absolute links, as opposed to a "pointer" relative path. If so, we don't want absolute links between folders in the unpacked folder, otherwise I don't think the asar will work on other computers.

test/cli-spec.js
```diff
@@ -188,4 +188,49 @@ describe('command line interface', function () {
...
+        fs.readlinkSync('tmp/packthis-with-symlink.asar.unpacked/Hello.framework/Headers'),
+        'Versions/Current/Headers',
+      );
+      assert.ok(
+        fs
+          .realpathSync('tmp/packthis-with-symlink.asar.unpacked/Hello.framework/Headers')
+          .endsWith('tmp/packthis-with-symlink.asar.unpacked/Hello.framework/Versions/A/Headers'),
+      );
+    } else {
+      assert.ok(
```
beyondkmp, line 229: After git cloning on Windows, SymlinkedDir is not a symlink but rather a regular file. I suggest only testing this on macOS/Linux. No need to test on Windows since you're not handling symlink in window.

https://github.com/mmaietta/asar/blob/1676a0ea070967f818167164be5bf7d0e378fb41/src/disk.ts#L65-L68

BlackHole1:
> Note: This implementation is isolated to mac/linux. It seems symlinks created on Windows (mklink) do not get properly committed to git, they seem to be committed as an actual file and I couldn't figure out why

I encountered this problem before when reviewing #308. Creating symlinks on Windows and uploading them to GitHub will cause CI to fail. I also tried creating symlinks in git bash, but it still didn't work. Perhaps Windows symlinks and Unix symlinks are completely incompatible? I'm not sure.

The method I used is to create them dynamically during testing, rather than using a fixtures-like solution. See: https://github.com/electron/asar/pull/308/files#diff-ab0f83364de04b521d00ee1ea7c4ecae6a5053e8e67588bc3c91cc468006c6cfR14-R28

mmaietta:
Thanks @BlackHole1 ! I'll get working right on that.

BlackHole1 (approved):
LGTM. Thanks!

PTAL @erickzhao

PR #343, opened 2024-11-21
test: remove unused index.test-d.ts file

dsanders11:
This isn't needed after #320, and nothing is running it since `tsd` was removed in that PR.


malept (approved):
Does what it says on the tin

PR #347, opened 2025-01-11
feat!: bump `engines` requirement to Node 22

erikian:
BREAKING CHANGE: bumps required Node.js version to `>=22.12.0`. ESM-only.


test/util/compareDirectories.js
```diff
@@ -1,14 +1,12 @@
-'use strict';
+import _ from 'lodash';
```
erickzhao, line 1: As a follow-up we could probably look into removing `lodash` altogether.

bin/asar.js
```diff
@@ -1,20 +1,21 @@
 #!/usr/bin/env node
```
erickzhao, line 1: nit: any preferences in `.js` vs `.mjs` for these JavaScript CLI scripts? I chose `.mjs` for `osx-sign`, but I guess either works as long as we have `type: module` in the package.json.

src/wrapped-fs.ts
```diff
@@ -22,17 +26,15 @@ type AsarFS = typeof import('fs') & {
...
+    (wrappedFs as any)[method] = fs[method];
   }
 }
 // To make it more like fs-extra
-promisified.mkdirp = (dir) => fs.promises.mkdir(dir, { recursive: true });
-promisified.mkdirpSync = (dir) => fs.mkdirSync(dir, { recursive: true });
-
-export default promisified;
+wrappedFs.mkdirp = (dir) => fs.promises.mkdir(dir, { recursive: true });
+wrappedFs.mkdirpSync = (dir) => fs.mkdirSync(dir, { recursive: true });
```
mmaietta, line 40: Not blocking, but potential improvement we could make here while we're at it.
```
export const wrappedFs: AsarFS = Object.keys(fs).reduce<AsarFS>((accum, method) => {
  return {
    ...accum,
    [method]: promisifiedMethods.includes(method) ? fs.promises[method] : fs[method]
  }
}, {
  mkdirp: (dir) => fs.promises.mkdir(dir, { recursive: true }),
  mkdirpSync: (dir) => fs.mkdirSync(dir, { recursive: true })
})
```

tsconfig.json
```diff
@@ -16,11 +11,9 @@
       "src/types"
     ],
     "allowSyntheticDefaultImports": true,
-    "moduleResolution": "node",
     "declaration": true,
     "noImplicitAny": true,
     "strictNullChecks": true,
```
mmaietta, line 16: Non-blocking, but can we add `"noUnusedLocals": true` while we're at it?

bin/asar.js
```diff
@@ -49,7 +50,7 @@ program.command('list <archive>')
     options = {
       isPack: options.isPack
     }
-    var files = asar.listPackage(archive, options)
+    var files = listPackage(archive, options)
     for (var i in files) {
       console.log(files[i])
     }
```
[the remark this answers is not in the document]
erikian: whoops - ece1eb04128dce47cae027bf95dd9a948bfa101

package.json
```diff
@@ -12,8 +13,9 @@
     "lib"
   ],
   "engines": {
-    "node": ">=10.12.0"
+    "node": ">=22.12.0"
   },
+  "engineStrict": true,
```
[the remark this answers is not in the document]
erikian: Yeah, this was an attempt to keep throwing errors when using unsupported Node versions without having to handle it in code. I tested with Node 3 / npm 2 and it did prevent a local copy of `@electron/asar` from being installed, but I didn't realize this behavior has long been removed.

I think we can just remove it and let the ESM syntax take care of ancient Node versions tbh -  e5e527dc71c8d07c61fc5b2bd2becb89ab87336e

package.json
```diff
@@ -40,20 +42,21 @@
...
+    "@types/mocha": "^10.0.10",
+    "@types/node": "~22.10.7",
     "chai": "^4.5.0",
-    "electron": "^22.0.0",
-    "electron-mocha": "^13.0.1",
+    "electron": "^34.5.0",
+    "electron-mocha": "^13.1.0",
     "lodash": "^4.17.15",
-    "mocha": "^10.1.0",
+    "mocha": "~11.0.2",
```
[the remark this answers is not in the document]
erikian: Yes, `mocha-chai-jest-snapshot` doesn't seem to play nice with `mocha@^11.1.0` and we get `Exception during run: e [YError]: .parseSync() must not be used with asynchronous builders, handlers, or middleware`. See https://github.com/mochiya98/mocha-chai-jest-snapshot/issues/25

package.json
```diff
@@ -40,20 +42,21 @@
   },
   "dependencies": {
     "commander": "^5.0.0",
-    "glob": "^7.1.6",
+    "glob": "^11.0.1",
     "minimatch": "^3.0.4"
```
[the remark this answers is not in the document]
erikian: `glob` actually traverses the filesystem, while `minimatch` only does string matching. Our tests say that the experimental [`path.matchesGlob`](https://nodejs.org/api/path.html#pathmatchesglobpath-pattern) method works exactly the same way, but since we're waiting for `fs.glob` to become stable we probably want to wait for `path.matchesGlob` as well.

`glob` does depend on `minimatch`, so at least we're getting it for free as long as we use `glob`; I bumped `minimatch` to the latest version so we don't ship two different versions - d1dd5eb714074038573c6c70249b7c437537e79

src/integrity.ts
```diff
@@ -1,7 +1,6 @@
-import * as crypto from 'crypto';
-import * as fs from 'fs';
-import * as stream from 'stream';
-import { promisify } from 'util';
+import crypto from 'node:crypto';
+import stream from 'node:stream';
+import { promisify } from 'node:util';
```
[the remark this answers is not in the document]
erikian: Ah nice catch, there were a couple of other `promisify` usages as well - 66d607a57f33920401f6950817cca00d443b1ce0

test/util/compareDirectories.js
erikian: That turned out to be quite straightforward using some `Set` methods that have just landed on Node 22 - fb66a5b5f4997dd22a0f74edf3c29f610c86ddfa

bin/asar.js
erikian: `.mjs` sounds good - 217fa7d461b93b547fd2c615f2e115c0ac370e81

src/wrapped-fs.ts
```diff
@@ -1,4 +1,8 @@
-const fs = 'electron' in process.versions ? require('original-fs') : require('fs');
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+
+const fs = 'electron' in process.versions ? require('original-fs') : require('node:fs');
```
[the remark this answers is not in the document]
erikian: I had to update `electron` to `^34.5.0` (Node 20.19.0) in our `devDependencies` because we run our Mocha tests on Electron through `electron-mocha` and [that's the first version](https://nodejs.org/en/blog/release/v20.19.0) where `require(esm)` works without the `--experimental-require-module` flag — we get `ERR_REQUIRE_ESM` on lower versions, though we could use Electron 32.1.0 (Node 20.17.0) and up by changing our `mocha` script to:

```json
"mocha": "NODE_OPTIONS=\"--experimental-require-module\" xvfb-maybe electron-mocha && mocha",
```

So to answer your question, I think Electron core should be able to upgrade to `@electron/asar@^4.0.0` as soon as Electron 36 is out and Electron 34 becomes the oldest supported version (it still uses Node 20.x — if we want to wait for Node 22.x we'd need to wait for Electron 37 to be released).

package.json
```diff
@@ -40,20 +42,21 @@
   },
   "dependencies": {
     "commander": "^5.0.0",
```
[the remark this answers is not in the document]
erikian: It only needs an `import` to be updated, so we can probably just do it in this PR - 0be538d057202ca1559562883b7c779cfeaac995

src/wrapped-fs.ts
erikian: 15062937edec4624030732bb27f195228477b4ed

tsconfig.json
erikian: 1c9bd61824e52dc4dc212ab89904a00032035600

dsanders11 (approved):

dsanders11:
@erikian, since this PR has been open for a while, can you merge in `main` to ensure tests still pass?

tsconfig.json
```diff
@@ -16,11 +11,10 @@
...
-    "esModuleInterop": true
+    "noUnusedLocals": true
```
erickzhao, line 17: nit: i think `@typescript-eslint/no-unused-vars` already covers this case?

README.md
```diff
@@ -16,7 +16,7 @@ all files together without compression, while having random access support.
 
 ### Install
 
-This module requires Node 10 or later.
+This module requires Node 22.12.0 or later.
```
erickzhao, line 19: blocking: we should update the usage examples below to reflect the new non-default exports and use ESM syntax.

erickzhao (changes requested):

erickzhao (approved):

georgexu99 (approved):

PR #350, opened 2025-01-27
feat: allow providing properties (such as `unpack`) in `ordering` input file

mmaietta:
Part 1 for fixing https://github.com/electron/universal/issues/117

This implementation achieves proposal `2b`
Excerpt below:
>Bypass minimatch with more explicit control over packaging options, similar as to asar ordering logic.
b. Leverage `options.ordering` configuration of electron/asar to pass in stringified JSON config for each file to be unpacked

In allowing more explicit control within the ordering file, advanced usage of electron/asar is provided (and kind of needed for electron/universal's current auto-unpack logic) to bypass limitations of the `minimatch` package

------

Right now, the "ordering" file is simply of format:
```
<file path>:<anything>
```
Right now, `electron/asar` only pulls the filename value before the `:`
https://github.com/electron/asar/blob/121efebd2d02e374713cb513eb56a0e3db9534e9/src/asar.ts#L92-L95

This proposal is to leverage `<anything>` to be `JSON.stringify` of Type:
```
{
	unpack: boolean
	// allows for future properties to be added for advanced usage without a breaking change
}
```
Notes:
- Retains the same logic as before for `--ordering` property, just extracted to a separate function
- Adds helper function to the symlinked test app "generator" to easily allow generating `ordering` file sets for any test that wants to leverage it.
- Taking this approach allows future configuration options to be allowed on a per-file basis should electron/asar need/desire to support more advanced usage

### DIFF
--- a/src/asar.ts
+++ b/src/asar.ts
@@ -80,19 +80,21 @@ async function getFileOrdering(
   const total = filenames.length;
 
   const fileOrderingSorted: FileProperties[] = [];
+  const isAlreadySorted = (file: string) =>
+    fileOrderingSorted.findIndex((config) => file === config.filepath) > -1;
 
-  ordering
-    .filter((config) => !fileOrderingSorted.includes(config) && filenames.includes(config.filepath))
-    .forEach((config) => {
+  for (const config of ordering) {
+    if (!isAlreadySorted(config.filepath) && filenames.includes(config.filepath)) {
       fileOrderingSorted.push(config);
-    });
+    }
+  }
 
-  filenames
-    .filter((file) => fileOrderingSorted.findIndex((config) => file === config.filepath) === -1)
-    .forEach((file) => {
+  for (const file of filenames) {
+    if (!isAlreadySorted(file)) {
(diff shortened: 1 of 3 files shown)


felixrieseberg (approved):
I had only one small comment, but I'm overall okay with this. Seems honestly like a good approach. I'll wait for @VerteDinde to take a look but otherwise I'm game for shipping this through!

src/asar.ts
```diff
@@ -33,6 +33,71 @@ function isUnpackedDir(dirPath: string, pattern: string, unpackDirs: string[]) {
...
+  let missing = 0;
+  const total = filenames.length;
+
+  const fileOrderingSorted: FileProperties[] = [];
+
+  ordering
+    .filter((config) => !fileOrderingSorted.includes(config) && filenames.includes(config.filepath))
+    .forEach((config) => {
+      fileOrderingSorted.push(config);
+    });
```
mmaietta, line 88: Just realized this `filter=>foreach` needs to be a `for-of`. Will push this fix asap

felixrieseberg (approved):

PR #351, opened 2025-01-28
test: Migrate to `vitest` with `.unpacked` and Asar Header snapshots

mmaietta:
Was having a fun chat with @erickzhao and wanted to see if I could lend a hand!

This PR accomplishes a few things:
- Migrates from `mocha` to `vitest`
- Converts all test files to typescript 🙌🏻
- Adds a new helper function `verifySmartUnpack` that I pulled from electron-builder's testing approach
  - It reads the asar header and validates it against the snapshot file
  - Provides a hook for additional asar `Filesystem` operations/verifications should so be desired
  - "Walks" the `unpacked` directory and stores a snapshot of the folder/file/symlink hierarchy.

With `verifySmartUnpack`, the header snapshot allows finer granularity of what is happening under the hood, especially with regards to symlinks and integrity calculations. Example respective snapshots excerpts below:
```
    "var": {
      "link": "private/var",
      "unpacked": true,
    },
```
```
"dir1": {
 "files": {
   "file1.txt": { 
     "integrity": {
       "algorithm": "SHA256",
       "blockSize": 4194304,
       "blocks": [
         "420149d3f852894ba7f32e9d3ec7d8919ee2451724bf2580b0186cd373bd6d82",
       ],
       "hash": "420149d3f852894ba7f32e9d3ec7d8919ee2451724bf2580b0186cd373bd6d82",
     },
     "size": 9,
     "unpacked": true,
   },
 },
}
```
Then the "walk" snapshot functionality allows all these lines to become just `verifySmartUnpack('tmp/packthis-unpack-subdir-cli.asar')`
https://github.com/electron/asar/blob/121efebd2d02e374713cb513eb56a0e3db9534e9/test/cli-spec.js#L182-L185

Additional Notes:
- `vitest` required dropping `rimraf.sync(...)` from the `beforeEach` hooks since the tests are in parallel and would cause random tests to fail. I've moved the `rimraf` to the `vitest-setup.ts`
- Ideally, to retain the previous functionality of `beforeEach`->`rimraf`, the tests need to run in isolation (which I wasn't able to achieve with [`singleThread`](https://vitest.dev/config/#pooloptions-threads-singlethread)) OR a random tmp dir is used each time that is unique to each test case. The latter option would require too large of a change and I wanted to keep changes in this PR isolated to the `test/` folder as much as possible.
- `src/asar.ts` file change is required though for `options?: ListOptions` to match its upstream `.listFiles(options)` API
- `lib` required being bumped to `es2019` so as to have access to `flat()` API
- cleaned up package.json `devDependencies`

Should this PR be accepted w/ `verifySmartUnpack` approach, the next steps would be cleaning up the test `asserts` that are now covered by the "walking" snapshots.


mmaietta:
Whelp, looks like this PR would depend on the upcoming node 22 upgrade. CC @erickzhao 
```
error vitest@3.0.4: The engine "node" is incompatible with this module. Expected version "^18.0.0 || ^20.0.0 || >=22.0.0". Got "16.20.1"

```

PR #357, opened 2025-02-12
test: add `--ordering` unit tests with `chai`

mmaietta:
(Proposing this as an alternative to migrating to jest?)
Note: This was test-only changes, but I was encouraged to add jsdoc to `ordering`, so I'm not sure if `test:` is the correct tag now.

Following up from a convo and the jest PR, this provides interim support for snapshot tests while still leveraging `mocha` and `electron-mocha`. This PR adds dynamically generated `ordering.txt` files for test coverage of current ordering logic

Adds 3 unit tests to cover ordering.txt files w/:
- format: `filepath`
- format: `: filepath`
- format: `random-number : filepath`

Functionally, this PR adds:
- GHA `test.yaml` -> `fail-fast: false` allows the matrix jobs to continue even when one fails. This will allow verifying asar behavior across each node version before the overall test is marked as a failure.
- `verifySmartUnpack` helper function that does the following:
   - `verifyHeader` - parses out unstable properties ("offset" varies per build machine?), converts all paths (e.g. `link`) to system independent strings, and then stores it as a snapshot
   - `verifyFileTree` - Traverses a directory and snapshots the tree hierarchy and plain text files (also used for unpacked dir snapshots)
- `createSymlinkedApp` - Creates a new app at runtime from the app template, adds folders, symlinks, and allows you to pass in `additionalFiles` that you can use to customize each app's directory

With `verifySmartUnpack`, the header snapshot allows finer granularity of what is happening under the hood, especially with regards to symlinks and integrity calculations. Example respective snapshots excerpts below:
```
    "var": {
      "link": "private/var",
      "unpacked": true,
    },
```
```
"dir1": {
 "files": {
   "file1.txt": { 
     "integrity": {
       "algorithm": "SHA256",
       "blockSize": 4194304,
       "blocks": [
         "420149d3f852894ba7f32e9d3ec7d8919ee2451724bf2580b0186cd373bd6d82",
       ],
       "hash": "420149d3f852894ba7f32e9d3ec7d8919ee2451724bf2580b0186cd373bd6d82",
     },
     "size": 9,
     "unpacked": true,
   },
 },
}
```
Then the "walk" snapshot functionality allows all these lines to become just included in `verifySmartUnpack('tmp/packthis-unpack-subdir-cli.asar')`
https://github.com/electron/asar/blob/121efebd2d02e374713cb513eb56a0e3db9534e9/test/cli-spec.js#L182-L185

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,9 @@
+module.exports = {
+    "watch-files": [
+       "test/**/*.js",
+       "lib/**/*.js"
+    ],
+    "recursive": true,
+    "file": "./mocha.setup.js", // setup file before everything else loads
+    "forbid-only": true
+}
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,9 +1,6 @@
 module.exports = {
-    "watch-files": [
-       "test/**/*.js",
-       "lib/**/*.js"
-    ],
-    "recursive": true,
-    "file": "./mocha.setup.js", // setup file before everything else loads
-    "forbid-only": true
-}
+  'watch-files': ['test/**/*.js', 'lib/**/*.js'],
+  recursive: true,
+  file: './mocha.setup.js', // setup file before everything else loads
+  'forbid-only': process.env.CI ?? false, // make sure no `test.only` is merged into `main`
+};
--- a/test/cli-spec.js
+++ b/test/cli-spec.js
@@ -194,13 +194,17 @@ describe('command line interface', function () {
   it('should unpack static framework with all underlying symlinks unpacked', async () => {
     const { testPath } = await createSymlinkApp('app');
     await execAsar(
-      `p ${testPath} tmp/packthis-with-symlink.asar --unpack *.txt --unpack-dir var --exclude-hidden`,
+      `p ${testPath} tmp/packthis-with-symlink1.asar --unpack *.txt --unpack-dir var --exclude-hidden`,
     );
 
-    await verifySmartUnpack('tmp/packthis-with-symlink.asar');
+    await verifySmartUnpack('tmp/packthis-with-symlink1.asar');
   });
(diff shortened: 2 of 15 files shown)


test/cli-spec.js
```diff
@@ -190,21 +192,53 @@ describe('command line interface', function () {
...
+      `p ${testPath} tmp/packthis-with-symlink.asar --ordering=${orderingPath} --exclude-hidden`,
+    );
+    await verifySmartUnpack('tmp/packthis-with-symlink.asar');
+  });
+  it('should respect ordering file (format: "${random number} : ${filepath}")', async () => {
+    const { testPath, filesOrdering } = await createSymlinkApp('app');
+
+    const orderingPath = path.join(testPath, '../ordered-app-ordering3.txt');
+    const data = filesOrdering.reduce((prev, curr) => {
+      return `${prev}${Math.floor(Math.random() * 1000)} :  ${curr} \n`;
```
erickzhao, line 235: question: what happens if the random number has a collision?

mmaietta: Nothing, current logic ignores anything before the `:`
https://github.com/electron/asar/blob/9b7ccfec05292d47ea82cb44cd45339bc058febc/src/asar.ts#L93-L95

.mocharc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    "watch-files": [
+       "test/**/*.js",
+       "lib/**/*.js"
+    ],
+    "recursive": true,
+    "file": "./mocha.setup.js", // setup file before everything else loads
+    "forbid-only": true
+}
```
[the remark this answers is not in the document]
mmaietta: I'll add root-level `./*.js` scripts to `prettier:write` as well

erickzhao (approved):

PR #358, opened 2025-03-07
feat: allow optional destination field for `asar extract-file` / `asar ef`

mmaietta:
Adds an optional parameter to `asar ef` using `[destination]` optional arg syntax

Resolves 2 `it.todo` unit tests with the support of an output directory

### DIFF
--- a/bin/asar.js
+++ b/bin/asar.js
@@ -55,13 +55,20 @@ program.command('list <archive>')
     }
   })
 
-program.command('extract-file <archive> <filename>')
-  .alias('ef')
-  .description('extract one file from archive')
-  .action(function (archive, filename) {
-    require('fs').writeFileSync(require('path').basename(filename),
-      asar.extractFile(archive, filename))
-  })
+  program
+  .command('extract-file <archive> <filename> [destination]')
+   .alias('ef')
+  .description(
+    'extract one file from archive. Optionally output to (already-existing) destination folder, otherwise cwd will be used',
+  )
+  .action(function (archive, filename, destination) {
+    const path = require('path');
+    const file = path.basename(filename);
+    destination = destination?.trim()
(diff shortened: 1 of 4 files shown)


bin/asar.js
```diff
@@ -55,13 +55,20 @@ program.command('list <archive>')
...
-  .description('extract one file from archive')
-  .action(function (archive, filename) {
-    require('fs').writeFileSync(require('path').basename(filename),
-      asar.extractFile(archive, filename))
-  })
+  program
+  .command('extract-file <archive> <filename> [destination]')
+   .alias('ef')
+  .description(
+    'extract one file from archive. Optionally output to (already-existing) destination folder, otherwise cwd will be used',
```
erickzhao, line 62: Should we do an `fs.existsSync` here to handle non existent paths more gracefully?

mmaietta: Great callout. Definitely can do! Okay with this error message to be thrown?

Related, what if we also throw an error if the destination file already exists? So that the user doesn't accidentally overwrite a file (like a sensitive/system resource if they enter a path with a space in it and don't wrap it in quotes lol?)
```
  .action(function (archive, filename, destination) {
    const path = require('path');
    const fs = require('fs');

    const file = path.basename(filename);
    destination = destination?.trim()
    const out = destination ? path.join(destination, file) : file;
    if (!fs.existsSync(destination)) {
      throw new Error("destination directory does not exist, please create before attempting extraction")
    }
    if (fs.existsSync(out)) {
      throw new Error("destination file already exists")
    }
    fs.writeFileSync(out, asar.extractFile(archive, filename));
  });
```

mmaietta: Updated the logic! Note: I had to `asar.extractFile` before `fs.existsSync` to retain previous behavior/unit test of "error thrown when file does not exist in asar"

Issue #344, opened 2024-12-22
No API to generate ASAR from in memory assets resources.


ChipZhang:
I have a project, which has a script to deal with ASAR format. In this proejct, I have generated compiled files suck as main.js, preload.js, ..., as well as the package.json, etc. Those files contents are in memory. I need to generate the ASAR archive based on those files and then send the archive to other prorgrams.

Currently, I have to create a tmp dir, write the assets files to the tmp dir, then use @electron/asar to package the ASAR archive and read this ASAR file. And finally delete all those stuff. This is very ugly and annoying.

As the official low level library to deal with ASAR format, you should provide APIs to pack/unpack/list ASAR purely in memory.

PR #360, opened 2025-03-27
feat: add in-memory API for streaming files to destination asar 

mmaietta:
feat: adding `async function createPackageFromStreams(dest: string, filestreams: Filestream[])` to allow streaming files directly from in-memory into the asar. Also will improve performance since files can be streamed instead of written temporarily to disk (as is done in electron-builder due to limitations of `minimatch`. Ref: https://github.com/electron/universal/issues/117)

Resolves: https://github.com/electron/asar/issues/344

Notes:
- The stream order is used as the insertion order.
- Consolidated logic for generating Pickle `out` `WriteableStream` and creating unpacked symlinks
- We use `streamGenerator: () => ReadStream` since we currently read the file multiple times (file data + file integrity). [This could probably use a separate refactor]
- Added unit test that takes a current fixture, creates a read stream of each file and sends that configuration to `createPackageFromStreams`
  - Verified that `asar extract` correctly contains all fixture data
  - Unit test verifies that the extracted contents exactly matches the src folder's contents

Related notes:
- I've been meaning to reorganize/refactor electron/asar to clean up duplicated logic and extract logic to isolated files per code intent.
   - I left out these optimizations to keep the diff of this PR on the "smaller" side
- This is somewhat the precursorary work to fixing https://github.com/electron/universal/issues/117 
- t's worth noting that forcing packagers to write files to the filesystem first, before creating an asar from that folder, seems to have significant performance impact, especially on Windows 11 Server specifically. The malware detection on Windows 11 Server significantly hampers the packaging time duration. (One might ask why not just use `createPackageFromFiles`, however, it does not allow transforms of file/folder hierarchy, and also runs into a separate issue when too many files are passed in; I think I recall due to the minimatch limitation mentioned further below but it's been a hot minute since I first explored that implementation logic flow)
- I think we could also leverage this new functionality to resolve an issue with electron/universal, where it currently tries to pass [a minimatch with 1.8k entries](https://github.com/electron/universal/issues/117) for a simple electron-quick-start project with an unpacked dependency. Instead of using a minimatch pattern for unpacked dirs+files, we could stream the necessary files to electron/asar with the unpacked boolean flag as part of the stream, bypassing [minimatch's innate limitation.](https://github.com/isaacs/minimatch/blob/0569cd3373408f9d701d3aab187b3f43a24a0db7/src/assert-valid-pattern.ts#L1) Note: I still need to spec/investigate this separately, but wanted to propose the initial streaming feature for electron/asar to solicit broader input/feedback.


mmaietta:
Confirming that this implementation is functional within electron-builder integration/refactor to use streams. GH runs in PR: https://github.com/electron-userland/electron-builder/pull/8995
Note: Docker containers aren't able to install a custom dependency of a branch/fork it seems, but Mac and Windows runners can? As such, I can only test electron-builder's test suite on mac/windows runners.

Confirmed app launches on Mac and properly works at runtime

Marking as Ready for Review

BlackHole1:
>  Note: Docker containers aren't able to install a custom dependency of a branch/fork it seems, but Mac and Windows runners can? As such, I can only test electron-builder's test suite on mac/windows runners.

I checked the code from PR https://github.com/electron-userland/electron-builder/pull/8995 and the reason for the CI failure.

The cause is related to yarn. You need to add `-e SKIP_YARN_COREPACK_CHECK=0` in `package.json -> scripts -> test-linux`:

```diff
-"test-linux": "docker run --rm -e CI=${CI:-true} -e DEBUG=${DEBUG:-} -e UPDATE_SNAPSHOT=${UPDATE_SNAPSHOT:-false} -e TEST_FILES=\"${TEST_FILES:-HoistedNodeModuleTest}\" -v $(pwd):/project -v $(pwd)-node-modules:/project/node_modules -v $HOME/Library/Caches/electron:/root/.cache/electron -v $HOME/Library/Caches/electron-builder:/root/.cache/electron-builder ${TEST_RUNNER_IMAGE_TAG:-electronuserland/builder:22-wine-mono} /bin/bash -c \"pnpm install && pnpm test-all\"",
+"test-linux": "docker run --rm -e CI=${CI:-true} -e DEBUG=${DEBUG:-} -e UPDATE_SNAPSHOT=${UPDATE_SNAPSHOT:-false} -e TEST_FILES=\"${TEST_FILES:-HoistedNodeModuleTest}\" -e SKIP_YARN_COREPACK_CHECK=0 -v $(pwd):/project -v $(pwd)-node-modules:/project/node_modules -v $HOME/Library/Caches/electron:/root/.cache/electron -v $HOME/Library/Caches/electron-builder:/root/.cache/electron-builder ${TEST_RUNNER_IMAGE_TAG:-electronuserland/builder:22-wine-mono} /bin/bash -c \"pnpm install && pnpm test-all\"",
```

Ref: https://github.com/yarnpkg/yarn/issues/9015

src/filesystem.ts
```diff
@@ -173,6 +169,13 @@ export class Filesystem {
     return link;
   }
 
+  private resolveLink(p: string) {
+    const symlink = fs.readlinkSync(p);
+    const parentPath = fs.realpathSync(path.dirname(p));
```
BlackHole1, line 174: Keep the comments to make it easier for other maintainers to understand, while also fixing issues with previously incorrect comments.

```suggestion
    // /var/tmp => /private/var
    const parentPath = fs.realpathSync(path.dirname(p));
```

BlackHole1 (approved):
Code LTGM

PR #371, opened 2025-08-04
build: use yarn@v4

MarshallOfSound:
Just seeing if this let's socket work a bit better


vitest.electron-reporter.ts
```diff
@@ -0,0 +1,39 @@
+import { RunnerTestFile } from "vitest"
+import { Reporter } from "vitest/reporters"
+
+export default class ElectronExitReporter implements Reporter {
```
erickzhao, line 4: Do we want to move this reporter into its own package for re-use?

package.json
```diff
@@ -47,16 +46,13 @@
   },
   "devDependencies": {
     "@tsconfig/node22": "^22.0.1",
-    "@types/mocha": "^10.0.10",
-    "@types/node": "~22.10.7",
-    "chai": "^5.2.0",
+    "@types/node": "^22.12.0",
```
erickzhao, line 49: We had initially pinned `"~22.10.7"` as the version range so that we wouldn't accidentally introduce new features from minor releases of Node.js that v22.12 didn't have

erickzhao (approved):

package.json
MarshallOfSound: I don't think `@types/node` tracks node versions like that? I could be wrong tho

PR #380, opened 2025-10-26
feat: add integrity digest management commands and functionality

nmggithub:
This is to add support for the additional integrity digest API I am adding in this pull request: https://github.com/electron/electron/pull/48587. I believe this package is the best place to put this functionality. Once merged, this functionality can be pulled into tools like [Electron Forge](https://github.com/electron/forge) as needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,11 @@ jobs:
           node-version: "${{ matrix.node-version }}"
           cache: 'yarn'
       - name: Install dependencies
-        run: yarn install --frozen-lockfile
+        run: yarn install --immutable
+      - name: Lint
+        run: yarn prettier:check
+      - name: Build
+        run: yarn build
       - name: Test
         env:
           ELECTRON_INTEGRITY_DIGEST_TEST_MAJOR_VERSION: '41'
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+yarn lint-staged
--- a/package.json
+++ b/package.json
@@ -33,11 +33,11 @@
     "test:watch": "vitest",
     "test:ui": "vitest --ui",
     "lint": "yarn prettier:check && yarn test:typecheck",
-    "prettier": "prettier \"src/**/*.ts\" \"test/**/*.ts\" \"*.js\"",
+    "prettier": "prettier \"src/**/*.{ts,cts,mts}\" \"test/**/*.{ts,cts,mts}\" \"*.{js,cjs,mjs}\"",
     "prettier:check": "yarn prettier --check",
     "prettier:write": "yarn prettier --write",
-    "prepare": "tsc",
-    "postinstall": "tsc"
+    "prepack": "yarn build",
+    "prepare": "husky"
   },
   "dependencies": {
     "commander": "^13.1.0",
@@ -53,7 +53,9 @@
     "@types/plist": "^3.0.5",
     "@types/semver": "^7.7.1",
     "electron": "^35.7.5",
-    "prettier": "^3.3.3",
+    "husky": "^9.1.7",
+    "lint-staged": "^16.2.7",
+    "prettier": "^3.8.1",
     "semver": "^7.7.4",
     "typedoc": "~0.25.13",
     "typescript": "^5.5.4",
@@ -65,5 +67,11 @@
     "electron": {
       "built": true
     }
+  },
+  "lint-staged": {
+    "*.{json,md}": "prettier --write --experimental-cli",
+    "*.{js,ts, cjs, mjs, cts, mts}": [
+      "prettier --write --experimental-cli"
+    ]
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -18,9 +18,11 @@ __metadata:
     commander: "npm:^13.1.0"
     electron: "npm:^35.7.5"
     glob: "npm:^13.0.2"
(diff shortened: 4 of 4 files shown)


src/integrity-digest.ts
```diff
@@ -0,0 +1,274 @@
...
+
+type IntegrityDigestV1 = IntegrityDigest<1, { sha256Digest: Buffer }>;
+
+type AnyIntegrityDigest = IntegrityDigestV1; // Extend this union type as new versions are added
+
+// Integrity digest calculation functions
+
+type AsarIntegrity = Record<string, Pick<FileRecord['integrity'], 'algorithm' | 'hash'>>;
+
+function calculateIntegrityDigestV1(asarIntegrity: AsarIntegrity): IntegrityDigestV1 {
```
erickzhao: nit: why is this put into a separate function?

bin/asar.mjs
```diff
@@ -71,6 +72,26 @@ program.command('extract <archive> <dest>')
     extractAll(archive, dest)
   })
 
+program.command('integrity-digest <app> <command>')
+  .alias('id')
+  .description('manage integrity digest in app binary')
```
erickzhao: Usage/docs should specify that this is macOS only.

question: do we want to stub out the CLI command on non-macOS platforms?

erickzhao (changes requested):

bin/asar.mjs
nmggithub: That's a fair point. And I think, as far as other platforms are concerned I think a simple error output would suffice. I think that can be done in a single function. Will add.

nmggithub: Actually, do our macOS app makers work on non-Apple platforms? If so, we might not want to limit this to only running on Apple platforms.

erickzhao (commented):
If it's possible, can we add tests to this PR as well?

src/integrity-digest.ts
```diff
@@ -0,0 +1,341 @@
...
+      throw new UnknownIntegrityDigestVersionError(version);
+  }
+}
+
+/**
+ * Gets the stored integrity digest for the app.
+ * @param appPath - The path to the app bundle.
+ * @returns The stored integrity digest for the app.
+ */
+export async function getStoredIntegrityDigestForApp<T extends AnyIntegrityDigest>(
```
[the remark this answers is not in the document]
nmggithub: Fixed as of 3173d04.

src/integrity-digest.ts
```diff
@@ -0,0 +1,382 @@
...
+}
+
+/// Integrity digest handling errors (API)
+
+export const InvalidAppPathError = class extends Error {
+  constructor() {
+    super('Invalid app path');
+    this.name = 'InvalidAppPathError';
+  }
+};
```
dsanders11:
```suggestion
export class InvalidAppPathError extends Error {
  constructor() {
    super('Invalid app path');
    this.name = 'InvalidAppPathError';
  }
}
```

Same with the errors below, unless there's a reason I'm missing on why they are currently done this way.

src/integrity-digest.ts
```diff
@@ -0,0 +1,382 @@
...
+  return {
+    used: true,
+    version: 1,
+    sha256Digest: integrityHash.digest(),
+  };
+}
+
+/**
+ * Reads the integrity information from the Info.plist of the given app
+ *  bundle and calculates the v1 integrity digest for the app.
```
dsanders11:
```suggestion
 * bundle and calculates the v1 integrity digest for the app.
```

src/integrity-digest.ts
nmggithub: Yeah, not sure why I missed this. Will fix!

src/integrity-digest.ts
nmggithub: I prefer adding additional leading whitespace if its a continuation of the above line. If this is against some style guide, please let me know!

test/integrity-digest-integration-spec.ts
```diff
@@ -0,0 +1,232 @@
+import path from 'node:path';
+import crypto from 'node:crypto';
+import { spawn } from 'node:child_process';
+import { promisify } from 'node:util';
```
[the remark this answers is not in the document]
nmggithub: Fixed in 8f27aaa.

test/integrity-digest-integration-spec.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    const baseDir = path.join(TEST_APPS_DIR, 'integrity-digest-integration');
+    const extractedDir = path.join(baseDir, 'extracted');
+
+    fs.rmSync(baseDir, { recursive: true, force: true });
+    await fs.mkdirp(baseDir);
+
+    const electronVersion = await getTargetElectronVersion();
+    const parsedElectronVersion = semver.parse(electronVersion);
+    if (!parsedElectronVersion || 41 > parsedElectronVersion.major) {
```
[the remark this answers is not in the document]
nmggithub: I chose the number-first pattern on purpose for style, but I can see your point here. The only issue is that your new logic does not handle a case where `parsedElectronVersion` is undefined. I _will_ make some changes to this, though.

src/integrity-digest.ts
dsanders11: I don't know that it's against any style guide we have, it just caught my eye. As long as it still renders the same in the type docs, I'm fine with it. 👍

PR #454, opened 2026-07-02
test: bump Electron and drop ElectronExitReporter

dsanders11:
`ElectronExitReporter` was an implicit workaround for an Electron issue that was fixed in electron/electron#51428 (the code comment here misattributes what the issue was), and now with E43 that fix is released and this workaround is no longer necessary.

The only remaining bit needed to get Vitest to run nicely inside of Electron is manually exiting on Vitest teardown as otherwise it waits for Node.js to exit naturally, but Electron stays running.


erickzhao (approved):
sweet :)

